PR #2, opened 2017-02-27
Initial version of UAST package

smola:


smola:
See this PR for the java-driver for an usage example:
https://github.com/bblfsh/java-driver/pull/2

uast/original.go
```diff
@@ -0,0 +1,276 @@
...
+	return nil
+}
+
+func toString(v interface{}) (string, error) {
+	switch o := v.(type) {
+	case string:
+		return o, nil
+	case fmt.Stringer:
+		return o.String(), nil
+	case int:
```
erizocosmico: if this is for printing the things inside a tree, all numbers are `int64` after decoding json with a `map[string]interface{}`. Also, `bool` and `float64` are valid values.

You can directly stringify them all with `fmt.Sprint`

smola: Done :+1:

uast/original.go
```diff
@@ -0,0 +1,241 @@
...
+	// topLevelIsRootNode is true if the top level object is the root node
+	// of the AST. If false, top level object should have a single key, that
+	// being the root node.
+	topLevelIsRootNode = false
+	// InternalRoleKey is a key string uses in properties to use the internal
+	// role of a node in the AST, if any.
+	InternalRoleKey = "internalRole"
+)
+
+// BaseOriginalToNoder is an implementation of OriginalToNoder that aims to work
```
ajnavarro: BaseOriginalToNoder->BaseToNoder

ajnavarro: OriginalToNoder maybe sould be ToNoder?

uast/original.go
```diff
@@ -0,0 +1,241 @@
...
+
+func (c *BaseToNoder) toNode(obj interface{}) (*Node, error) {
+	m, ok := obj.(map[string]interface{})
+	if !ok {
+		return nil, ErrUnexpectedObject.New("map[string]interface{}", obj)
+	}
+
+	n := NewNode()
+	for k, o := range m {
+
```
ajnavarro: space.

uast/original.go
```diff
@@ -0,0 +1,241 @@
...
+	if len(src) > 1 {
+		return nil, ErrUnexpectedObjectSize.New(1, len(src))
+	}
+
+	for _, obj := range src {
+		return c.toNode(obj)
+	}
+
+	panic("not reachable")
```
ajnavarro: I don't like this. In my opinion I could rewrite this to:
``` go
src, ok := v.(map[string]interface{})
if !ok {
	return nil, ErrUnsupported.New("non-object root node")
}

if topLevelIsRootNode {
	return nil, ErrUnsupported.New("top level object as root node")
}
 
 if len(src) > 1 {
 	return nil, ErrUnexpectedObjectSize.New(1, len(src))
 }

var node *Node 
for _, obj := range src {
	node = c.toNode(obj)
        break
}

if node == nil {
       return nil, ErrEmptyAST.New()
}

return node, nil
```

uast/original.go
```diff
@@ -0,0 +1,241 @@
...
+
+			n.Children = append(n.Children, children...)
+		default:
+			if err := c.addProperty(n, k, o); err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	sort.Sort(byOffset(n.Children))
```
ajnavarro: missed space.

uast/original.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+func (c *BaseToNoder) mapToNode(k string, obj map[string]interface{}) (*Node, error) {
+	n, err := c.toNode(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	n.Properties[InternalRoleKey] = k
+	return n, nil
```
ajnavarro: missed space.

uast/original.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	return ns, nil
+}
+
+func (c *BaseToNoder) addProperty(n *Node, k string, o interface{}) error {
+	switch {
+	case c.isTokenKey(k):
+		if n.Token != "" {
+			return fmt.Errorf("two token keys for same node: %s", k)
```
ajnavarro: I think we can use a specific `var Err...` using "srcd.works/go-errors.v0"

uast/original.go
```diff
@@ -0,0 +1,241 @@
...
+	case c.InternalTypeKey == k:
+		s := fmt.Sprint(o)
+		if err := c.setInternalKey(n, s); err != nil {
+			return err
+		}
+
+		tk := c.syntheticToken(s)
+		if tk != "" {
+			if n.Token != "" {
+				return fmt.Errorf("two token keys for same node: %s", k)
```
ajnavarro: I think we can use a specific `var Err...` using "srcd.works/go-errors.v0"

uast/original.go
erizocosmico: +1 to avoiding unreachable panics.

uast/original.go
erizocosmico: If we want the sorts to be always the same, we should use `sort.Stable` instead of `sort.Sort`.

uast/original.go
erizocosmico: Are we using go-errors in here? If we are, +1 to that

uast/original.go
```diff
@@ -0,0 +1,241 @@
...
+func toUint32(v interface{}) (uint32, error) {
+	switch o := v.(type) {
+	case string:
+		i, err := strconv.ParseUint(o, 10, 32)
+		if err != nil {
+			return 0, err
+		}
+
+		return uint32(i), nil
+	case uint32:
```
erizocosmico: if this comes from a JSON-parsed `map[string]interface{}` the only possibility here is `int64` because that's the only integer type JSON handles when it does not know the type.

`map[string]interface{}`, `[]interface{}`, `string`, `bool`, `int64` and `float64` are the only types that can come out of a JSON unmarshalled to `map[string]interface{}`.

uast/rules.go
```diff
@@ -0,0 +1,119 @@
+package uast
+
+type Rule func(...*Node) error
```
erizocosmico: these types should have docs

smola:
@erizocosmico @ajnavarro rebased, added your requested changes and some executable examples.

PR #6, opened 2017-03-01
Make sure go-bindata is installed before using it.

Serabe:


Serabe:
@mcuadros @smola

Serabe:
Did not see #4

PR #20, opened 2017-03-06
uast: new conversion dsl

smola:
* uast/ann: moved to package uast/ann
* uast/ann: improve path matching
* uast/ann: DSL is now more fluent
* uast/ann: added HasToken, Not
* uast: add NodePath
* uast: PreOrderVisit not supports ErrNoVisit to avoid visiting part of a tree.


uast/ann/ann_test.go
```diff
@@ -0,0 +1,251 @@
+package ann
+
+import (
+	"testing"
+
+	. "github.com/bblfsh/sdk/uast"
+
+	"fmt"
```
ajnavarro: bad import group

smola: done

uast/uast.go
```diff
@@ -326,3 +326,37 @@ const (
 func (f IncludeFlag) Is(of IncludeFlag) bool {
 	return f&of != 0
 }
+
+type NodePath []*Node
```
alcortesm: We have to explain that the first node is supposed to be the root and the last one one is supposed to be the target or destination of the path.  Depending of the name chosen to explain this, maybe we should change the name of the Node method for that name.

uast/uast.go
```diff
@@ -326,3 +326,37 @@ const (
...
+
+func (p NodePath) Ancestors() []*Node {
+	if len(p) <= 1 {
+		return nil
+	}
+
+	return p[:len(p)-1]
+}
+
+func (p NodePath) ParentPath() NodePath {
```
smola: IsRoot and Ancestors removed, since they were not used.

uast/uast.go
smola: Added comments, renamed to Path.

uast/ann/ann.go
```diff
@@ -0,0 +1,219 @@
+// Package ann provides a DSL to annotate UAST.
+package ann
+
+import (
+	"github.com/bblfsh/sdk/uast"
+)
+
+// PathMatcher is a function that matches against a NodePath. If a prefix is not
+// matched, it returns that prefix. If there is no match at all, it returns the
+// input NodePath. If it matches the full NodePath, nil is returned.
```
smola: Clarified PathMatcher godoc.

PathMatcher is an interface that can be implemented by anything. PathPredicate is a convenient way of signaling that something is a predicate that takes path, and as such, it implements the PathMatcher interface.

It is also implemented by NodePredicate.

Implementing PathMatcher interface directly in the predicates allows a more fluent DSL.

For example:
```
On(HasInternalType("root")).Role(File)
	Rules(
		On(Any(), HasInternalType("import")).Roles(ImportDeclaration),
		On(Not(HasInternalType("import"))).Roles(TypeDeclaration),
	)
```

`On` works on any PathMatcher, while `Not` works on NodePredicate. Implementing the PathMatcher interface in NodePredicate allows combining them fluently.

Any type change is welcome, but keep in mind that this package is a DSL oriented to make writing UAST annotators in a declarative way easily.

PR #23, opened 2017-03-07
add driver implementation

smola:


protocol/drivermain.go
```diff
@@ -0,0 +1,176 @@
+package protocol
```
mcuadros: maybe instead will be best `github.com/bblfsh/sdk/protocol/driver/cmd/`

mcuadros: and the main in the driver, following the same pattern of sdk cmd

protocol/internal/testdriver/main.go
```diff
@@ -0,0 +1,15 @@
+package main
```
mcuadros: why internal?

smola: `testdriver` package is used only in tests, this `main.go` then can be used (with a simple replacement) in `etc/skeleton/driver/main.go.tpl`.

protocol/drivermain.go
smola: Ok. Note that this does not build a command itself, but it's used from driver's main.go.

smola:
This PR depends on https://github.com/bblfsh/sdk/pull/24

PR #31, opened 2017-03-21
Added ColumnKey to BaseToNoder and load its value into (existing) Node.StartPosition.Col

juanjux:
Python AST uses line and column, the later was missing from BaseToNoder but was present in the StartPosition struct member of the Node.

### DIFF
--- a/uast/tonoder.go
+++ b/uast/tonoder.go
@@ -43,7 +43,7 @@ type BaseToNoder struct {
 	OffsetKey string
 	// LineKey is a key that indicates the line number.
 	LineKey string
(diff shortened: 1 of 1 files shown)


uast/tonoder.go
```diff
@@ -43,6 +43,8 @@ type BaseToNoder struct {
 	OffsetKey string
 	// LineKey is a key that indicates the line number.
 	LineKey string
+	// ColumnKey is a key that indicated the column inside the line
```
smola: s/indicated/indicates

PR #38, opened 2017-03-23
uast: add Static role

smola:


smola:
Please, do not merge this. Annotating the `static` modifier itself as `Static` is the wrong approach.

PR #40, opened 2017-03-27
uast: add Call roles

smola:


smola:
@juanjux For variadic functions,  there is no difference in the call, only in the declaration.

juanjux:
@smola true, I messed up with the definition, in Python you still can pass *args, and **kwargs to function calls but these are not specific of function calls but a special syntax to "decompress" lists or dictionaries (I will make a separate PR for those).

PR #43, opened 2017-03-28
uast: add character and boolean literals, add literals doc

smola:
At the moment, this covers literals for Java (boolean, character, null, number, string, type).


juanjux:
Also, there is "ByteLiteral" that depending on the interpretation of "Character" of the language could be needed to differentiate from encoded chars.

smola:
@juanjux I added some of the ones you mentioned. For other cases, I would like to see a few examples for multiple languages.

One of the things that does not fit these annotations are Go's composite literals (e.g. struct literals).

juanjux:
@smola I was thinking more about C99/C++ struct literals that are definition and instantiation at the same time than Go struct literals that AFAIK are more like just instantiation, but they certainly are a very specific case and could be added just as an instantiation.

PR #46, opened 2017-03-29
uast: added most common binary comparison operators

juanjux:
These are the ones than Python AST's use but I guess most languages have the same. The negated or XEqual could be composed by two operators but from what I'm seeing languages usually use different nodes for them.


uast/uast.go
```diff
@@ -30,6 +30,18 @@ const (
 	Expression
 	Statement
 
+	// Binary comparison operators (usually >, >=, etc)
+	BinaryCompOpEqual
```
smola: Maybe just `OpEqual`?

juanjux: The equal sign is usually (ab)used on many things in most languages (setting default values, assigment, augmented assigment, start of the method in Nim, etc) I think using OpCompEqual indicates more explicitly that this is the equal sign when used in comparison/test expressions (i. e. == in Cish languages).

uast/uast.go
```diff
@@ -30,6 +30,41 @@ const (
 	Expression
 	Statement
 
+	// Comparison operators. Usually inside one of the Condition nodes, but could also be in an
+	// an expression of boolean value by itself. The tested expressions will be its siblings.
+
+	// OpCompEquality is the operator that tests for logical equality between two expressions
+	OpCompEquality
```
smola: Maybe simplify to `OpEqual`, `OpNotEqual`, `OpLessThan`, `OpGreaterThan`, `OpenLessThanOrEqual`,  `OpGreaterThanOrEqual`, `OpSame` (instead of Identity), `OpNotSame` ?

`OpCompPertenency` maybe `OpContains`?

juanjux: Yeah, Boolean and Bitwise need their qualifiers to avoid conflicts but comparison operators doesn't really need it. Will update the PR.

PR #47, opened 2017-03-29
uast: added boolean operators

juanjux:
No need to explain I guess :)


uast/uast.go
```diff
@@ -33,6 +33,12 @@ const (
 	// File is the root node of a single file AST.
 	File
 
+	// BooleanOperators
```
smola: Section names in comments, if any, should not go together with a constant. Otherwise, it shows up as the documentation of that constant.

juanjux:
Ok, I will fix this and the other PRs

uast/uast.go
```diff
@@ -33,6 +33,17 @@ const (
 	// File is the root node of a single file AST.
 	File
 
+	// Binary boolean operators, like
+
+	// OpBooleanAnd is the boolean AND operator (i.e. "and" or && in most languages)
+	OpBooleanAnd
+	// OpBooleanAnd is the boolean OR operator (i.e. "or" or || in most languages)
```
smola: s/OpBooleanAnd/OpBooleanOr
etc

PR #48, opened 2017-03-29
uast: added bitwise operators

juanjux:
I added the [C ones](en.wikipedia.org/wiki/Bitwise_operations_in_C) that are probably the most standard. Will do a separate PR for the assignOperator ones.


uast/uast.go
```diff
@@ -27,6 +27,20 @@ const (
 	// or more qualifiers.
 	QualifiedIdentifier
 
+	// Binary bitwise operators. Examples from C syntax given.
+	// <<
+	BinaryBitwiseOpLeftShift
```
smola: `BinaryBitwise -> Bitwise` ?

PR #49, opened 2017-03-29
uast: added arithmetic operators

juanjux:
Basic ones. I didn't add some more exotic ones like pow (** in Python, ^ in other languages) or the two or more kinds of division but we could discuss if we need to add them or add an "BinaryAritmeticOther" with an annotation.

### DIFF
--- a/uast/uast.go
+++ b/uast/uast.go
@@ -94,6 +94,21 @@ const (
 	// OpBooleanXor is the boolean XOR operator (i.e. "XOR" or ^ in most languages)
 	OpBooleanXor
 
+	// Binary aritmethic operators. Examples with C operators.
+	// TODO: should we have division and FloorDivision like Python or Nim?
+	// TODO: should we had the pow operator that some languages have?
+
+	// OpAdd is the binary add operator (i.e. + in most languages).
+	OpAdd
+	// OpSubstract is the binary subtract operator (i.e. - in most languages).
+	OpSubstract
+	// OpMultiply is the binary multiply operator (i.e. * in most languages).
+	OpMultiply
+	// OpDivide is the binary division operator (i.e. / in most languages).
+	OpDivide
+	// OpMod is the binary division module operator (i.e. % or "mod" in most languages).
+	OpMod
+
 	// PackageDeclaration identifies the package that all its children
 	// belong to. Its children include, at least, QualifiedIdentifier or
 	// SimpleIdentifier with the package name.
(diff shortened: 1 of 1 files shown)


smola:
@juanjux Names are misleading. Maybe it should be `OpAdd`, `OpSub`, etc?

Then bitwise could be named as `OpBitwiseAnd`, boolean could be `OpBooleanAnd`, set ops could be `OpSetIntersection`, etc.

juanjux:
@smola mmm I don't think this is misleading, these roles would be explicitly for the operators in aritmetic expressions, boolean and bitwise would also be explicit (I like explicit things myself).

There is however the matter of languages that allow you to overload operators (C++, Python, D), in that case it could or could not be logically "aritmetic" but for example Python mark it as such even when overloaded.

uast/uast.go
```diff
@@ -33,6 +33,21 @@ const (
 	// File is the root node of a single file AST.
 	File
 
+	// Binary aritmethic operators. Examples with C operators.
+	// TODO: should we have division and FloorDivision like Python or Nim?
+	// TODO: should we had the pow operator that some languages have?
+
+	// OpArithmeticAdd is the binary add operator (i.e. + in most languages)
```
alcortesm: End all comments with a full stop, they are ordinary English phrases that will be used verbatim in the documentation so they must obey standard English language rules.

uast/uast.go
```diff
@@ -33,6 +33,21 @@ const (
...
+	OpArithmeticAdd
+	// OpArithmeticSubstract is the binary subtract operator (i.e. - in most languages)
+	OpArithmeticSubstract
+	// OpArithmeticMultiply is the binary multiply operator (i.e. * in most languages)
+	OpArithmeticMultiply
+	// OpArithmeticDivision is the binary division operator (i.e. / in most languages)
+	OpArithmeticDivision
```
alcortesm: we probably need 3 "division" operators instead of division and mod:
- Division
- Quotient
- Remainer

alcortesm: There are two ways to call these things:

|  Action       |  Name          |
| ------------- |-------------|
| Add      | Addition |
| Subtract | Subtraction |
| Multiply | Multiplication |
| Divide | Division |

Right now we are mixing both, we should choose one.

uast/uast.go
juanjux: Fixed.

uast/uast.go
juanjux: True, I'll change it. I will also drop the "Arithmetic" part of the name since there is not conflict with other operators (unlike that happens with boolean/bitwise operators.)

juanjux: About the division: I decided to went with what most languages I checked use for division and remainder in their AST node namings. For example as I wrote in the "TODO", Python3 also have // for integer division. Quotient would just be the one of the two subnode literal/symbol/expressions, we could add it as a role but I don't think is really needed.

PR #51, opened 2017-03-29
uast: unary operators

juanjux:
I think this cover the most important unary operators. There is also the option of composing them using the Binary[Aritmetic|Bitwise]Operators (changing their name to just operators) but some like takeaddress or dereference doesn't have analogue ones there.


uast/uast.go
```diff
@@ -30,6 +30,28 @@ const (
 	Expression
 	Statement
 
+	// Unary operators
+	// ++x
+	UnaryPreIncrement
```
smola: Maybe it's better to use naming: such as `OpPreIncrement`?

uast/uast.go
```diff
@@ -30,6 +30,28 @@ const (
...
+	// x++
+	UnaryPostIncrement
+	// --x
+	UnaryPreSubstract
```
smola: `Substract` could be replaced with `Decrement`, to not confuse it with binary substraction operation.

uast/uast.go
```diff
@@ -30,6 +30,36 @@ const (
...
+	// Unary operators. These will have a single child node that will apply the operator over it.
+	// TODO: use BooleanNot to implement the UnaryNot?
+
+	// OpPreIncrement increments in place the value before it is evaluated. It's
+	// typical of C-inspired languages (e. g. ++x).
+	OpPreIncrement
+	// OpPostIncrement increments in place the value after it is evaluated. It's
+	// typical of C-inspired languages (e. g. x++).
+	OpPostIncrement
+	// OpPreIncrement decrement in place the value before it is evaluated. It's
```
smola: s/OpPreIncrement/OpPreDecrement

juanjux:
Fixed.

PR #55, opened 2017-03-29
uast: Delete keyword

juanjux:
This is the delete keyword that unallocates stuff created on the heap and call its destructors on some languages (Python, C++, D, Java, etc). Python also uses it for removing keys from dictionaries.


juanjux:

PR #56, opened 2017-03-29
uast: Yield (generators) and YieldFrom (coroutines)

juanjux:
Yield is common on most modern languages, used to implement generators. Yield From (used to implement cooperative coroutines) is less common as a built in keyword because on most languages that's implemented on libraries or its trivial to implement using other language built-in features (haskell continuations) -- so maybe we could add Yield From as an annotation to Yield.


juanjux:

PR #57, opened 2017-03-29
uast: added OpUnpackList and OpUnpackMap

juanjux:
This maps the * and ** features of Python (and many other, usually interpreted, languages) to unpack a list or map into anywhere that expects several comma or k=v-comma separated values (function calls usually).


juanjux:

PR #58, opened 2017-03-29
uast: BlockScopeResourceExpr and BlockScopeResourceAlias

juanjux:
For languages that support a special syntax for RAII'ing resources inside a block like Python "with stuff do". It could also serve for Java "Try with resources" tagging the TryExpr node with and adittional BlockScopeResourceExpr (or adding a child, but I think using multiple roles where it makes sense makes the UAST more concise and easy to human-parse).

### DIFF
--- a/uast/uast.go
+++ b/uast/uast.go
@@ -290,10 +290,25 @@ const (
 	// Block is a group of statements. If the source language has block scope,
 	// it should be annotated both with Block and BlockScope.
 	Block
-	// BlockScope is a block with its own block scope.
+	// BlockScope is a block with its own block scope. Depending on the language it could
+	// have children of the following types.
 	// TODO: Should we replace BlockScope with a more general Scope role that
 	//       can be combined with Block?
 	BlockScope
+	// BlockScopeName is the given name of a scoped block for languages that support it
+	// (e.g. C++ namespaces).
+	BlockScopeName
+	// BlockScopeBody is the body holding the expressions inside the BlockScope.
+	BlockScopeBody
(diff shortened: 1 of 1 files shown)


juanjux:
New commit, added more BlockScope related roles:

- BlockScopeName (will be needed for C++ namespaces)
- BlockScopeBody (since BlockScopes with this PR could have several childs this would hold the expressions in the proper block scope).

uast/uast.go
```diff
@@ -189,10 +189,25 @@ const (
...
 	BlockScope
+	// BlockScopeName is the given name of a scoped block for languages that support it
+	// (e.g. C++ namespaces).
+	BlockScopeName
+	// BlockScopeBody is the body holding the expressions inside the BlockScope.
+	BlockScopeBody
+	// BlockScopeResourceExpr marks the role of the expression that will
+	// generate a resource that will be cleaned when the block goes out
+	// of scope, RAII-style (e.g. Python's "with open('file.txt') as f:" or
+	// Java "Try with resources" that combines this with a Try...Catch
```
alcortesm: A full stop is missing here at the end.

uast/uast.go
```diff
@@ -189,10 +189,25 @@ const (
...
+	BlockScopeResourceExpr
+	// This is an alias for a BlockScopeResourceExpr that can be used
+	// inside the BlockScope (e.g. the "f" in Python's: "with open('file.txt') as f:")
```
alcortesm: Add a full stop at the end.

juanjux: Both fixed, thanks!

juanjux:

PR #59, opened 2017-03-30
UAST: Index operator and slice roles (+ friends)

juanjux:
This is a complex one, so feel free to tag it as needs-research; I've looked at a lot of languages and I think these roles should covert most of the indexing operations except some exotic ones like Python's "stride" third slice element.

Please review and comment if you feel the need since these roles are very important to get right.

http://rosettacode.org/wiki/Assigning_Values_to_an_Array

http://rosettacode.org/wiki/Array

### DIFF
--- a/uast/uast.go
+++ b/uast/uast.go
@@ -112,6 +112,45 @@ const (
 	// stored in a pointer or reference type (e. g. &x).
 	OpTakeAddress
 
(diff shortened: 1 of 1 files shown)


juanjux:
PS: I'm removing OpIndexAssign. Even tough some languages have it in their AST as separate nodes, it can just be composed as an Assignment node with an OpIndex as the lhs.

juanjux:

PR #62, opened 2017-03-31
uast: function declaration additional nodes

juanjux:
This adds nodes that are almost universal:

FunctionDeclarationBody
FunctionDeclarationName
FunctionDeclarationReceiver
FunctionDeclarationArgument
FunctionDeclarationArgumentName
FunctionDeclarationArgumentDefaultValue (this one is not so universal but still pretty common)
FunctionDeclarationVarArgsList (I'll add VarArgsMap on the next slightly less universal PR)


uast/uast.go
```diff
@@ -152,8 +152,37 @@ const (
 	// in a certain scope.
 	ImportAlias
 
-	// TODO: arguments, return value, body, etc.
+	// TODO: argument type declarations, return value, body, etc.
+
+	// FunctionDeclaration is the parent node of all function or method declarations. It should have
+	// FunctionDeclarationName and FunctionDeclarationBody children and, if the function has formal arguments,
+	// FunctionDeclarationArgument children. Please note that some languages could support more than one
+	// FunctionDeclarationBody roles.
```
smola: `Please note that some languages could support more than one
FunctionDeclarationBody roles.`

I wouldn't add this here unless if have a good reason to do it.

Cases such as Elixir or Scala functions will require to carefully look at their AST to see how can we annotate them.

uast/uast.go
```diff
@@ -152,8 +152,37 @@ const (
...
 	FunctionDeclaration
+	// FunctionDeclarationBody is the grouping node for all nodes in the function body. Please note that some
+	// languages could support more than one FunctionDeclarationBody roles.
```
smola: `Please note that some
// languages could support more than one FunctionDeclarationBody roles.`

same here

juanjux:
@smola done. Removed that comment and added another one about possible FunctionDeclarations without FunctionBody nodes (C header file declarations, forward declarations in D, etc).

Issue #61, opened 2017-03-31
uast: merge roles for Assignment and AugmentedAssignment


smola:
Currently, we have:

```
	// Assignment represents a variable assignment or binding.
	// The variable that is being assigned to is annotated with the
	// AssignmentVariable role, while the value is annotated with
	// AssignmentValue.
	Assignment
	AssignmentVariable
	AssignmentValue

	// AugmentedAssignment is an augmented assignment usually combining the equal operator with
	// another one (e. g. +=, -=, *=, etc). It is expected that children contains an
	// AugmentedAssignmentOperator with a child or aditional role for the specific Bitwise or
	// Arithmetic operator used. The AugmentedAssignmentVariable and AugmentedAssignmentValue roles
	// have the same meaning than in Assignment.
	AugmentedAssignment
	AugmentedAssignmentOperator
	AugmentedAssignmentVariable
	AugmentedAssignmentValue
```

This feels quite redundant. I wonder if we can come up with a more succint way of annotating augmented assignments.

juanjux:
I will do a PR to remove AugmentedAssignmentVariable and AugmentedAssignmentValue since we can just use the Assignment* ones which have the same meanings. I'm not so sure about removing the AugmentedAssignment and AugmentedOperatorRoot; the first one have a first level representation on the languages I checked.

juanjux:
This is outdated because of #167.

PR #64, opened 2017-03-31
uast: augmented assignment roles simplified

juanjux:
Removes AugmentedAssignmentVariable and AugmentedAssignmentValue and documents to use the normal Assignment versions. From #61 .


smola:
Let's discuss next week about this.
See https://github.com/bblfsh/sdk/issues/61

juanjux:
I did this because of #61 :smile: but ok let's talk about these together.

uast/role_string.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by "stringer -type=Role"; DO NOT EDIT.
+// Code generated by "stringer -type=Role"; DO NOT EDIT
```
ajnavarro: :rofl:

ajnavarro: Maybe is better keep this file as it is.

juanjux: I've not the mininal idea of how that could have changed but it will be rewritten on the next go generate anyway

juanjux:

PR #74, opened 2017-04-05
uast: function declaration arguments parent role

juanjux:
Micro PR, it adds a grouping node (FunctionDeclarationArguments) for function formal parameters (FunctionDeclarationArgument). I did a quick check on Python, C++ (cdt) and Nim and all those create a parent node for the formal parameters.

Also updated the doc so its clearer what the singular and plural roles are for.


juanjux:
I'll also close and redo this PR since it also had the history pretty messed up.

PR #75, opened 2017-04-05
sdk tonoder: listproperty to parent node

juanjux:
This PR adds two settings to allow the translation to promote properties (dictionary keys) with a list value to its own node (aka "property-lists"). Without this PR the nodes in the list will be added as children of the enclosing dictionary with the only reference to the key being the internalRole property so if we have:

```text
{If: { body: [expr, expr, expr],
     test: [exprtest, exprtest],  
     else: [elseexpr, elseexpr]}
}
```

In the UAST we would get:

```text
{If: {Children: [expr, expr, expr, exprtest, exprtest, elseexpr, elseexpr]}}
```
This can work in some cases since the nodes can have multiple roles so a human/machine reading the UAST could still see the type of every nodes, so in this case expr would have the roles Expression and IfBody and the args would have Argument and Arguments, but it's a little different for what most AST do.

This PR allows the UAST translator to define that some of all of these property-lists can be upgraded to have its own node with the values of the list as its Children like:

```text
{If { body: {Children:[expr, expr, expr], 
     test: [exprtest, exprtest], 
     else: [exprelse, exprelse]
}}
```

In this case only the "body" would need to have the IfBody role, the test the IfCondition, etc, tough of course the normalizer programmer could change this as fit.

Please review not only the concept but also the code since I'm still very green to Go lang. I'm also usually bad at naming things so check that too.

Tests are not included tough I've tested it troughfully with my version of the Python driver but I'll add them on this same branch soon.

This sets the internal key of the new promoted nodes to "Parentnode.property" so the new nodes can have rules defined in the normalizer.

### DIFF
--- a/uast/tonoder.go
+++ b/uast/tonoder.go
@@ -50,6 +50,15 @@ type BaseToNoder struct {
 	// SyntheticTokens is a map of InternalType to string used to add
 	// synthetic tokens to nodes depending on its InternalType.
 	SyntheticTokens map[string]string
+	// PromotedPropertyLists allows to convert some properties in the native AST with a list value
+	// to its own node with the list elements as children. The key of the first map is the name
+	// of the InternalType which can have promotions and the value is a map where keys must be the names
+	// of the properties to be promoted if the value is true. For example to promote a "body" property
+	// inside an "If" InternalKey the map should contain: ["If"]["body"] = true.
+	PromotedPropertyLists map[string]map[string]bool
+	// If this option is set, all properties mapped to a list will be promoted to its own node. Setting
+	// this option to true will ignore the PromotedPropertyLists settings.
+	PromoteAllPropertyLists bool
 }
 
 func (c *BaseToNoder) ToNode(v interface{}) (*Node, error) {
@@ -86,7 +95,32 @@ func (c *BaseToNoder) toNode(obj interface{}) (*Node, error) {
 	}
 
 	n := NewNode()
-	for k, o := range m {
+
+	// We need to have the internalkey before iterating others
+	internalKey, err := c.getInternalKeyFromObject(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	var promotedKeys map[string]bool
+	if (!c.PromoteAllPropertyLists && c.PromotedPropertyLists != nil) {
+		promotedKeys = c.PromotedPropertyLists[internalKey]
+	}
+
+	if err := c.setInternalKey(n, internalKey); err != nil {
+		return nil, err
+	}
+
+	// Sort the keys of the map so the integration tests that currently do a
+	// textual diff doesn't fail because of sort order
+	var keys []string
+	for listkey := range m {
+		keys = append(keys, listkey)
+	}
+	sort.Strings(keys)
+
+	for _, k := range keys {
+		o := m[k]
 		switch ov := o.(type) {
 		case map[string]interface{}:
 			child, err := c.mapToNode(k, ov)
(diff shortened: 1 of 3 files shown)


uast/tonoder.go
```diff
@@ -50,6 +50,15 @@ type BaseToNoder struct {
 	// SyntheticTokens is a map of InternalType to string used to add
 	// synthetic tokens to nodes depending on its InternalType.
 	SyntheticTokens map[string]string
+	// PromotedPropertyLists allows to convert some properties in the native AST with a list value
+	// to its own node with the list elements as childs. The key of the first map is the name
```
smola: s/childs/children

uast/tonoder.go
```diff
@@ -50,6 +50,15 @@ type BaseToNoder struct {
...
+	// of the InternalType which can have promotions and the value is a map where keys must be the names
+	// of the properties to be promoted if the value is true. For example to promote a "body" property
+	// inside an If internal Key the map should contain: ["If"]["body"] = true
```
smola: s/internal Key/internal key/

or

s/internal Key/InternalKey/

juanjux:
@smola typos fixed and two tests added.

juanjux:
Closing for the moment because I found an odd bug, will reopen once I've fixed it.

juanjux:
Reopened; the problem was elsewhere.

uast/tonoder.go
```diff
@@ -50,6 +50,15 @@ type BaseToNoder struct {
 	// SyntheticTokens is a map of InternalType to string used to add
 	// synthetic tokens to nodes depending on its InternalType.
 	SyntheticTokens map[string]string
+	// PromotedPropertyLists allows to convert some properties in the native AST with a list value
+	// to its own node with the list elements as children. The key of the first map is the name
+	// of the InternalType which can have promotions and the value is a map where keys must be the names
+	// of the properties to be promoted if the value is true. For example to promote a "body" property
+	// inside an "If" InternalKey the map should contain: ["If"]["body"] = true
```
alcortesm: End the paragraph with a full stop.

uast/tonoder.go
```diff
@@ -50,6 +50,15 @@ type BaseToNoder struct {
...
+	PromotedPropertyLists map[string]map[string]bool
+	// If this option is set, all properties mapped to a list will be promoted to its own node. Setting
+	// this option to true will ignore the PromotedPropertyLists settings
```
alcortesm: End the paragraph with a full stop.

juanjux:
@alcortesm  fixed, thanks!

smola:
@juanjux Please, rebase and squash into a single commit.

juanjux:
@smola done.

uast/tonoder.go
```diff
@@ -86,6 +95,22 @@ func (c *BaseToNoder) toNode(obj interface{}) (*Node, error) {
...
+
+	// We need to have the internalkey before iterating others
+	internalKey, err := c.getInternalKeyFromObject(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	var promotedKeys map[string]bool
+	if (!c.PromoteAllPropertyLists) {
+		promotedKeys = c.PromotedPropertyLists[internalKey]
```
smola: If `c.PromotedPropertyLists` is nil this will fail, right? It shouldn't.

juanjux: Actually I tested  (and is nil on the first unittest) and:

- First access on the parent map when is nil: returns a "nil" map[string]bool
- Second access on this one when is nil (the line above): returns the default boolean value, false.

Check this:
https://play.golang.org/p/J4U7ipI7vn

But adding a nil check is not a bad idea anyway, I'll also add an aditional unittest because I just saw that I don't have a case for PromotedPropertyLists = nil and PromoteAllPropertyLists = true.

juanjux: @smola changes commited (squashed).

juanjux:
I also merged the bugfix about the sorting of the generated uast properties in this PR.

PR #81, opened 2017-04-10
uast: Added ComprehensionExpression nodes

juanjux:
This PR adds the UAST roles needed to map Python (and probably other languages) comprehension expressions (list/tuple, map and set comprehensions). These are syntactically complex but these nodes should covert at least the comprehensions as used in Python and Nim.

Probably should be tagged needs-research.

### DIFF
--- a/uast/uast.go
+++ b/uast/uast.go
@@ -405,6 +405,32 @@ const (
 	// (e.g. Visual Basic).
 	This
 
+	// List/Map/Set comprehensions are expressions that iterate over a container, optionally
+	// using restricting conditions, and generate a new List/Dict/Set with a subset of the
+	// elements of the original container optionally with some transformation applied to
+	// every element. These are parent nodes where the expression sublements will be mapped
+	// on the Comprehension* roles. The parent node is [List|Map|Set]ComprehensionExpression.
+	ListComprehensionExpression
+	// Parent node for Map comprehensions.
+	MapComprehensionExpression
+	// Parent node for Set comprehensions.
+	SetComprehensionExpression
+	// This is the parent node of all the loops contained inside the comprehension. Sometimes
+	// comprehensions can contain more than one loop to iterate over sub-containers inside a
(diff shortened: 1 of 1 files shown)


uast/uast.go
```diff
@@ -405,6 +405,34 @@ const (
 	// (e.g. Visual Basic).
 	This
 
+	// List/Map/Set comprehensions are expressions that iterate over a container, optionally
+	// using restricting conditions, and generate a new List/Dict/Set with a subset of the
+	// elements of the original container optionally with some transformation applied to
+	// every element. These are parent nodes where the expression sublements will be mapped
+	// on the Comprehension* roles.
```
alcortesm: This comment will probably be out of place in the auto-generated documentation as it is not just before any declaration.  Should be merged with other comments for it to work correctly.

uast/uast.go
```diff
@@ -405,6 +405,34 @@ const (
...
+
+	// Parent node for List comprehensions
```
alcortesm: Full stop at the end.

uast/uast.go
```diff
@@ -405,6 +405,34 @@ const (
...
+	ListComprehensionExpression
+	// Parent node for Map comprehensions
```
alcortesm: Full stop at the end.

uast/uast.go
```diff
@@ -405,6 +405,34 @@ const (
...
+	MapComprehensionExpression
+	// Parent node for Set comprehensions
```
alcortesm: Full stop at the end.

uast/uast.go
juanjux: Fixed.

uast/uast.go
juanjux: Thank you. Man, I should write a Vim plugin or something to catch these, I don't know why this mania of mine of skipping final dots in comments that translates to generated documentation in this case,.

juanjux:

PR #82, opened 2017-04-10
uast: added MapKey and MapValue roles

juanjux:
This complements the MapLiteral UAST role adding MapKey and MapValue.

### DIFF
--- a/uast/uast.go
+++ b/uast/uast.go
@@ -377,6 +377,12 @@ const (
 	// annotations.
 	OtherLiteral
 
+	// MapEntry is the expression pairing a map key and a value usually on MapLiteral expressions. It must
+	// have both a MapKey and a MapValue children (e.g. {"key": "value", "otherkey": "otherval"} in Python).
+	MapEntry
+	MapKey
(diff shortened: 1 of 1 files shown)


uast/uast.go
```diff
@@ -377,6 +377,12 @@ const (
 	// annotations.
 	OtherLiteral
 
+	// MapKeyValue is the expression pairing a map key and a value usually on MapLiteral expressions. It must
+	// have both a MapKey and a MapValue children (e.g. {"key": "value", "otherkey": "otherval"} in Python).
+	MapKeyValue
```
alcortesm: Setting the precedent of naming things by enumerating their contents is dangerous.  May we can call this MapEntry, MapEntryLiteral instead.

juanjux: You are right - I'm bad at naming things that this shows. What about MapEntry? So it will serve in other instances where is used to access keys and values and not only for literals.

alcortesm: I think MapEntry is quite good.

PR #85, opened 2017-04-18
remove leftover test, not needed anymore

alcortesm:
The tests were failing because this test references an absolute path in the docker container that does not exist when you clone the project (the path points to files outside the repository).

This patch removes the test entirely.


bzz:
CI is failing here, same as in #87 and #89 - I wonder, does modified binary asserts should be committed as well or not?

smola:
Oh, well, this fails because `make code-generate` was not executed. I merged without CI passing, so my fault here.

PR #87, opened 2017-04-18
bblfsh-tools do not fail inside a container, started from macOS

bzz:
`make test` pass, but though the process `bblfsh-tools` fail to run properly inside the container, when started from macOS host.

```
+ docker run --rm -t -u bblfsh:501 -v /Users/alex/go/src/github.com/bblfsh/java-driver:/opt/driver/src/ -v /Users/alex/go:/go -e ENVIRONMENT=bblfsh/java-driver-build-with-go bblfsh/java-driver-build-with-go make test-driver-internal
/go/bin/bblfsh-tools: line 1: ����: not found
/go/bin/bblfsh-tools: line 2: H__PAGEZEROx__TEXT�__text__TEXT: not found
/go/bin/bblfsh-tools: line 1: syntax error: unexpected end of file (expecting ")")
cd native; \
mvn package
Warning: JAVA_HOME environment variable is not set.
[INFO] Scanning for projects...
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] Building javaparser 1.0-SNAPSHOT
[INFO] ------------------------------------------------------------------------
```

As `bblfsh-tools` are used from Makefile in both, inside and outside of a build container, but are compiled once under the host architecture - it results in obscure failure to run MarchO binary on Linux. 

After more testing, this seems to not fail `make test` though, so not sure if this fix worth merging.


etc/build/Makefile
```diff
@@ -12,7 +12,13 @@ $(shell mkdir -p $(tmplocation))
 
 bblfsh-sdk := $(shell command -v bblfsh-sdk 2> /dev/null)
 bblfsh-tools := $(shell command -v bblfsh-tools 2> /dev/null)
-ifdef bblfsh-tools
+in-container := $(shell echo $$ENVIRONMENT)
+
+# TODO(bzz) ideal: `go run` ONLY inside container, started from non-linux host
+ifdef bblfsh-tools # with Golang
```
smola: Maybe this outer ifdef is not needed?

etc/build/Makefile
```diff
@@ -12,7 +12,13 @@ $(shell mkdir -p $(tmplocation))
 
 bblfsh-sdk := $(shell command -v bblfsh-sdk 2> /dev/null)
 bblfsh-tools := $(shell command -v bblfsh-tools 2> /dev/null)
-ifdef bblfsh-tools
+in-container := $(shell echo $$ENVIRONMENT)
+
+# TODO(bzz) ideal: `go run` ONLY inside container, started from non-linux host
```
smola:
```
$ file ~/dev/go/bin/bblfsh-tools 
/home/smola/dev/go/bin/bblfsh-tools: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, not stripped
```

Maybe you can check if it's non-linux host if `/go/bin/bblfsh-tools` is not an ELF 64-bit binary?

etc/build/Makefile
bzz: At first I also thought so, but it looks like this was here before, together with `command -v` for a reason: to avoid running this in native build container WITHOUT go.

etc/build/Makefile
smola: Sounds good.

etc/build/Makefile
smola: Alright.

etc/build/Makefile
bzz: addressed in 8f3e7570365a9115becbb916edf7c18823732638

bzz:
All feedback addressed, `make test` pass local on both, macOS and linux.

Ping @smola for review.

CI is failing with 

```
generated bindata is out of sync
make: *** [validate-commit] Error 2
The command "make validate-commit" failed and exited with 2 during .
```

smola:
@bzz Just run `make` or `make bindata` and update the commit with changes.

bzz:
@smola thanks, rebased on master and re-wrote the history with `make all`. Results is the same as in #98

`make validate-commit` passes locally but fails on CI :/

```
 var version string
generated bindata is out of sync
make: *** [validate-commit] Error 2
The command "make validate-commit" failed and exited with 2 during .
```

smola:
@bzz something went wrong when executing `make bindata`:

```
--- a/assets/skeleton/bindata.go

+++ b/assets/skeleton/bindata.go

@@ -978,7 +978,7 @@ var _driverMainGoTpl = []byte(`package main

 import (

 	"github.com/bblfsh/sdk/protocol"

 

-	"github.com/bblfsh/sdk/etc/skeleton/driver/normalizer" //REPLACE:"github.com/bblfsh/{{.Manifest.Language}}-driver/driver/normalizer"

+	"github.com/bblfsh/{{.Manifest.Language}}-driver/driver/normalizer"

 )
```

it seems the `driver-tpl` rule in Makefile didn't work. Do you have `sed` installed?

bzz:
Awesome, yes I do! :) It is just a GNU sed that is required here, and with BSD one it seems to fail.
I forgot that sed is not part of coreutils ..., so there are 3 thins for SDK now


- (macOS) for `mktemp`

   ```
   brew install coreutills
   echo 'export PATH="/usr/local/opt/coreutils/libexec/gnubin:$PATH"' >> ~/.bash_profile`
   ```
- (macOS) for `envsubst`

   ```
   brew install gettext
   echo 'export PATH="/usr/local/opt/gettext/bin:$PATH"' >> ~/.bash_profile`
   ```

- (macOS) for `sed`

   ```
   brew install gnu-sed
   echo 'export PATH="/usr/local/opt/gnu-sed/libexec/gnubin:$PATH"' >> ~/.bash_profile`
   ```


Do you think this list is worth putting in https://github.com/bblfsh/sdk/blob/master/README.md or is there a better place? (it's for SDK devs)

smola:
@bzz Does full build work with this PR?

bzz:
I'm a bit confused, what do you mean by full build?
Tested it on linux and macOS, both worked for me and CI passes.

PR #89, opened 2017-04-19
Update of README template of the drivers

bzz:
Small style fix + UAST converter dependencies requirement

### DIFF
--- a/etc/skeleton/README.md.tpl
+++ b/etc/skeleton/README.md.tpl
@@ -16,12 +16,13 @@ Development Environment
 Requirements:
 - `docker`
 - [`bblfsh-sdk`](https://github.com/bblfsh/sdk) _(go get -u github.com/bblfsh/sdk/...)_
+- UAST converter dependencies _(go get -t -v ./...)_
 
 To initialize the build system execute: `bblfsh-sdk prepare-build`, at the root of the project. This will install the SDK at `.sdk` for this driver.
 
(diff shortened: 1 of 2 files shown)


bzz:
CI fails with

```
generated bindata is out of sync
make: *** [validate-commit] Error 2
The command "make validate-commit" failed and exited with 2 during .
```

It looks like I need to commit generated resources as well.

bzz:
History re-written

bzz:
Rebased on latest masted and assets re-generated. 

`make validate-commit` passes locally for me :/

bzz:
Rebased on latest master, regenerated binaries, same as before, CI is failing, same as in #87

bzz:
CI passes now. Ping @smola for review

etc/skeleton/README.md.tpl
```diff
@@ -16,12 +16,13 @@ Development Environment
 Requirements:
 - `docker`
 - [`bblfsh-sdk`](https://github.com/bblfsh/sdk) _(go get -u github.com/bblfsh/sdk/...)_
+- UAST converter dependencies _(go get -t -v ./...)_
```
smola: actually this should be covered by `go get -u github.com/bblfsh/sdk/...`? What's the missing dependency?

bzz: That would be awesome, but for now [stretchr/testify](https://github.com/stretchr/testify) is missing so `mvn test` fails \wo it.

PR #91, opened 2017-04-26
Support for Infix/Prefix/Postfix order

juanjux:
This PR adds the roles Infix/Prefix and Postfix. These are intended to be used as additional roles to
tag the parents of expressions that come using one of these specific notations (Prefix will continue being the default). 

I've renamed the preOrderPathIter type and related method to just "orderPathIter" since it will switch from one iteration mode to another dynamically when it finds one of these roles. I also tried to implement different types for different iteration methods but it was getting really complex since the iterator would have to switch its own type to other inside the Next() method (and return back after the parsing of the tagged node and their children) which the Go language doesn't make easy to do without too many workarounds (no real polymorphism). So I think this solution is simpler, easier to understand and a better match for the language.

It also adds unittests.

Please note that this adds a breaking change to drivers since it renamed the NewPreOrderPathIter to NewOrderPathIter, of the public API, but the Python driver only used it in one unittest (once fixed the Python driver integration tests work).

Please review.


uast/iter.go
```diff
@@ -43,47 +43,105 @@ func (i *sliceIter) Next() Path {
...
+const (
+	preOrder = iota
+	inOrder
+	postOrder
+)
+
+func getNextIterType(n *Node) int {
+	order := preOrder
+	for _, r := range n.Roles {
+		if r == Infix {
```
alcortesm: this should be a switch statement.

juanjux: Definitely. Change already done, thanks.

uast/uast.go
```diff
@@ -40,6 +40,13 @@ const (
 	BinaryExpressionRight
 	BinaryExpressionOp
 
+	// Infix should mark the nodes with are parents of expression nodes using infix notation, e.g.: a+b.
```
smola: s/with are/which are/

uast/uast.go
```diff
@@ -40,6 +40,13 @@ const (
...
+	Infix
+	// Prefix should mark the nodes with are parents of expression nodes using prefix notation, e.g.: +ab.
+	Prefix
```
smola: I don't think we need the Prefix role at the moment, since that's default, so adding the role is just redundant.

uast/iter.go
```diff
@@ -43,47 +43,104 @@ func (i *sliceIter) Next() Path {
...
+		// Right
+		i.stack = append(i.stack, newNodeSliceIter(i.last, n.Children[1]))
+		// Relator
+		i.stack = append(i.stack, newNodeSliceIter(i.last, noChildsNodeCopy(n)))
+		// left
+		i.stack = append(i.stack, newNodeSliceIter(i.last, n.Children[0]))
+
+		i.last = i.stack[len(i.stack)-1].Next()
+	} else if iterType == postOrder && hasChilds {
+		// Childs
```
smola: s/Childs/Children/

uast/iter.go
```diff
@@ -43,47 +43,104 @@ func (i *sliceIter) Next() Path {
...
+}
 
+// Adds to the orderPathIter stack with the right order depending on
+// the order Role with (if set) can be Infix, Postfix or Prefix. Defaults to Preorder
+// if the order Role is not set. This also updates i.last.
+func (i *orderPathIter) addToStackWithOrder(n *Node) {
+	hasChilds := len(n.Children) > 0
+
+	iterType := getNextIterType(n)
+	if iterType == inOrder && hasChilds {
```
smola: s/hasChilds/hasChildren/

uast/iter.go
```diff
@@ -43,47 +43,104 @@ func (i *sliceIter) Next() Path {
...
 
-// NewPreOrderPathIter creates an iterator that iterates all tree nodes in pre-order.
-func NewPreOrderPathIter(p Path) PathStepIter {
-	return &preOrderPathIter{
+type orderPathIter struct {
+	stack []PathIter
+	last  Path
+}
+
+// NewOrderPathIter creates an iterator that iterates all tree nodes in pre-order.
```
smola: Update documentation, it is not in pre-order anymore.

uast/uast_test.go
```diff
@@ -30,3 +30,118 @@ func TestNodeTokens(t *testing.T) {
...
+			{InternalType: "right", Token: "tok_post_right"},
+			{InternalType: "op_postfix", Roles: []Role{Postfix}, Token: "Postfix+"},
+			}}
+	result := Tokens(n)
+	expected := []string{"id2", "tok_post_left", "tok_post_right", "Postfix+"}
+	require.Equal(expected, result)
+}
+
+// Test for mixed order roles
+func TestOrderTokens(t *testing.T) {
```
smola: You should add cases for which the left and right side of a prefix/infix/postfix is a tree with further tokens.

juanjux: :+1:  to all. I should really make the mouse give me an electrical discharge every time I write "childs". I'll add the test with further tokens as a separate case if you don't mind because I like the unittests to be as atomic as possible.

juanjux:
@smola committed all your requests and also fixed a slippery bug that showed on the new tests you requested.

smola:
@juanjux Please, double check. I don't see changes for my previous comments.

juanjux:
@smola WTF, I must have deleted the middle commit instead of squash... oh well, should be fixed now.

PR #101, opened 2017-05-09
UAST: added Goto role

juanjux:
We forgot to add the grandpa!

(this commits just add the role to the list and the regenerated go gen uast.go).


smola:
OMG!
:shipit:

PR #102, opened 2017-05-10
Added function for cyclomatic complexity + unittest

juanjux:
This adds uast.utils.CyclomaticComplexity that for a given Node (and children) will calculate the cyclomatic complexity. Also adds a unittest.

This doesn't include the NPath complexity.

### DIFF
--- a/uast/uast_test.go
+++ b/uast/uast_test.go
@@ -9,7 +9,7 @@ import (
 func TestPrefixTokens(t *testing.T) {
 	require := require.New(t)
 
-	n := &Node{InternalType: "empty",
+	n := &Node{InternalType: "module",
 		Children: []*Node{
 			{InternalType: "id", Token: "id3"},
 			// Prefix is the default so it doesnt need any role
@@ -25,7 +25,7 @@ func TestPrefixTokens(t *testing.T) {
 func TestPrefixTokensSubtree(t *testing.T) {
 	require := require.New(t)
 
-	n := &Node{InternalType: "empty",
+	n := &Node{InternalType: "module",
 		Children: []*Node{
 			{InternalType: "id", Token: "id3"},
 			// Prefix is the default so it doesnt need any role
@@ -53,7 +53,7 @@ func TestPrefixTokensSubtree(t *testing.T) {
 func TestPrefixTokensPlain(t *testing.T) {
 	require := require.New(t)
 
-	n := &Node{InternalType: "empty",
+	n := &Node{InternalType: "module",
 		Children: []*Node{
 			{InternalType: "id", Token: "id3"},
 			// Prefix is the default so it doesnt need any role
@@ -68,7 +68,7 @@ func TestPrefixTokensPlain(t *testing.T) {
 
 func TestInfixTokens(t *testing.T) {
 	require := require.New(t)
-	n := &Node{InternalType: "empty",
+	n := &Node{InternalType: "module",
 		Children: []*Node{
 			{InternalType: "id", Token: "id1"},
 			{InternalType: "op_infix", Roles: []Role{Infix}, Token: "Infix+", Children: []*Node{
@@ -83,7 +83,7 @@ func TestInfixTokens(t *testing.T) {
 func TestInfixTokensSubtree(t *testing.T) {
 	require := require.New(t)
 
-	n := &Node{InternalType: "empty",
+	n := &Node{InternalType: "module",
 		Children: []*Node{
 			{InternalType: "id3", Token: "id3"},
 			// Prefix is the default so it doesnt need any role
@@ -110,7 +110,7 @@ func TestInfixTokensSubtree(t *testing.T) {
 
 func TestInfixTokensPlain(t *testing.T) {
 	require := require.New(t)
-	n := &Node{InternalType: "empty",
+	n := &Node{InternalType: "module",
 		Children: []*Node{
 			{InternalType: "id", Token: "id1"},
 			{InternalType: "left", Token: "tok_in_left"},
@@ -124,7 +124,7 @@ func TestInfixTokensPlain(t *testing.T) {
 
 func TestPostfixTokens(t *testing.T) {
 	require := require.New(t)
-	n := &Node{InternalType: "empty",
+	n := &Node{InternalType: "module",
 		Children: []*Node{
 			{InternalType: "id", Token: "id2"},
 			{InternalType: "op_postfix", Roles: []Role{Postfix}, Token: "Postfix+", Children: []*Node{
@@ -139,7 +139,7 @@ func TestPostfixTokens(t *testing.T) {
 func TestPostfixTokensSubtree(t *testing.T) {
 	require := require.New(t)
 
-	n := &Node{InternalType: "empty",
+	n := &Node{InternalType: "module",
 		Children: []*Node{
 			{InternalType: "id", Token: "id2"},
 			// Prefix is the default so it doesnt need any role
@@ -165,7 +165,7 @@ func TestPostfixTokensSubtree(t *testing.T) {
 
 func TestPostfixTokensPlain(t *testing.T) {
 	require := require.New(t)
-	n := &Node{InternalType: "empty",
+	n := &Node{InternalType: "module",
 		Children: []*Node{
 			{InternalType: "id", Token: "id2"},
 			{InternalType: "left", Token: "tok_post_left"},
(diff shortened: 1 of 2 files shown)


uast/utils.go
```diff
@@ -17,3 +17,57 @@ func Tokens(n *Node) []string {
...
+			case Continue:
+				fallthrough
+			// FIXME: uncomment when Goto is added to the UAST
+			//case Goto:
+				//fallthrough
+			case OpBooleanAnd:
+				fallthrough
+			case OpBooleanOr:
+				fallthrough
+			case OpBooleanXor:
```
alcortesm: I think we can improve readability by merging all these cases into a single one:
```
switch(r) {
case If, SwitchCase, SwitchDefault,
         For, ForEach, While, DoWhile,
        ...
        OpBooleanXor:
        return complx
}
```
And then maybe consider if an if would be better than a switch if we are not planning on doing any other action different from complx++ on other cases.

uast/uast_test.go
```diff
@@ -217,3 +217,27 @@ func TestOrderTokens(t *testing.T) {
 		"id3", "Prefix+", "tok_pre_left", "subright_in1", "tok_pre_right", "subright_in2"}
 	require.Equal(expected, result)
 }
+
+func TestCyclomaticComplexity(t *testing.T) {
+	require := require.New(t)
+	n := &Node{InternalType: "empty",
```
alcortesm: I am sorry, I am a little bit out of shape on this, can you explain why an "empty" node is required? (no need to explain that on the code, just in this thread will be enough)

uast/utils.go
juanjux: I like your switch more. The if would be a little verbose maybe. I will modify the code and update the PR.

uast/uast_test.go
juanjux: Its just a parent node, the internalType is just some string for having some field that can be queried for debugging but certainly that one should be something like "module". I will change the code and update the PR.

juanjux: @alcortesm done.

uast/utils.go
juanjux: @alcortesm done.

uast/utils.go
```diff
@@ -17,3 +17,34 @@ func Tokens(n *Node) []string {
 	}
 	return tokens
 }
+
+// This returns the cyclomatic complexity for the node. This uses the method of counting one + one
```
smola: Godoc for functions start with the name of the function: `CyclomaticComplexity returns the cyclomatic complexity for the node...`.

juanjux: Yes, checked with a Python snippet that is semantically the same as the node in the unittest, I've also checked some implementations of this and they basically do the same ([example1](https://github.com/fzipp/gocyclo/blob/master/gocyclo.go#L215) [example2](http://checkstyle.sourceforge.net/config_metrics.html#CyclomaticComplexity)).

juanjux: Updated the PR with the doc fix.

uast/utils.go
```diff
@@ -17,3 +17,34 @@ func Tokens(n *Node) []string {
...
+
+	for {
+		p := iter.Next()
+		if p.IsEmpty() {
+			break
+		}
+		n := p.Node()
+		for _, r := range n.Roles {
+			switch(r) {
+			case If, SwitchCase, SwitchDefault, For, ForEach, While,
```
abeaumont: Goto role is missing.
Not sure if others (i. e. IfElse, Break, Continue, Return, Assert, Throw...) should be considered too.
Boolean operators won't necessarily be binary (e. g. and/or operators tend to have an arbitrary number of operands in Lisp languages), not sure if that matters.

juanjux: There are several definitions, I went with the most common. IfElse/Break/Continue and exceptions are almost never considered (I was surprised about IfElse too) , Return never and the switch "default" clause is probably 50%. Goto is definitely needed. The boolean operators are definitely not binary in lots of languages as you say, even in some that use infix (like Python) use lists in the AST, I think I've a PR about that. What is also important is to know if the boolean evaluation on tests is done on short circuit because that definitely  increases the complexity (one path for operator vs one path for the test). I accounted them as short circuit since its the most common option but we should update once we have some way to represent these language invariants in the AST or metadata.

abeaumont: It'd useful to add a reference to the definition we use. I looked at this to check how it should be calculated in Java, which cites those extra tokens: https://www.leepoint.net/principles_and_practices/complexity/complexity-java-method.html

About boolean operators, I'd say they short circuit in most if not all languages (it doesn't make much sense otherwise). I'm not sure I understand what you mean, should all the operands be counted for cyclomatic complexity or not? Current implementation doesn't seem to count them.

juanjux: That one is definitely exotic (wrong?) because almost no implementation consider the elses (in the original paper there are some snippets where the elses are not counted). For example check the comments for Java on: 

https://docs.sonarqube.org/display/SONAR/Metrics+-+Complexity

Please notice that the UAST's "IfElse" is not like a Python "elif" but just en "else"; is called IfElse because is a subnode of "If" following the nomenclature that we're using for other tokens like "ForUpdate" but in this case is somewhat confusing (and "else if" would be a "If" token child from an "ElIf" token).

I updated the PR and added an extended comment with several links and the original paper and clarifications on the different measures.

juanjux: PS: in the future we could add a CustomCycComplexity where the used can provide the tokens to consider to avoid more bikeshedding about the versions of this or, even better, a tree to cyclic graph converter that, among other things, could provide a real measure of the complexity.

abeaumont: Yes, it seems that else roles should not be considered. The link you provided seems to be rather inconsistent, in Java it doesn't consider throw, return, break, continue, etc. while they consider them in C/C++ (and according to the original paper, they should be considered).

About the boolean operators, I agree what the proposed complexity, but I don't understand why it's not implemented, do you mean that these operator tokens cannot have children? If so, why?

abeaumont: It'd be nice to settle which definition are we going to use, to avoid ambiguities. Are we going to use the original McCabe paper, any other particular implementation?

juanjux: The original McCabe paper only defines it with regard to an cyclic graph and only have a couple examples, that's why there are many interpretations.

abeaumont: So, which one should we be using? Maybe PMD's one http://pmd.sourceforge.net/pmd-4.3.0/xref/net/sourceforge/pmd/rules/CyclomaticComplexity.html ? Any other?

juanjux:
I guess than SonarQube tries to go with the variant more common on each language.

The real spec of the calculation we do is in the comment:

// If | SwitchCase |  SwitchDefault | For[Each] | [Do]While | TryCatch | Continue | OpBoolean* | Goto

I think I found it on MSDN but if you want to put a name to it we can search choose any of the ones that are the same (for example the SonarQube.Java one that is already linked in the comment). If you prefer other spec, do a PR, I don't really care if the rules are documented :)

uast/utils.go
```diff
@@ -18,7 +18,12 @@ func Tokens(n *Node) []string {
 	return tokens
 }
 
-// CyclomaticComplexity returns the cyclomatic complexity for the node. This uses the method of
+// CyclomaticComplexity returns the cyclomatic complexity for the node. The cyclomatic complexity
+// is a quantitative measure of the number of linearly independent paths through a program's source code.
+// It was developed by Thomas J. McCabe, Sr. in 1976. For a formal description see:
+// https://en.wikipedia.org/wiki/Cyclomatic_complexity
+// And the original paper: http://www.literateprogramming.com/mccabe.pdf
+// This implementation uses the method of
```
abeaumont: I'd be happier with a rationale about what Cyclomatic Complexity definition we chose and why:
- What Cyclomatic Complexity definition are we based on (I understood that we weren't using McCabe's paper definition, maybe I understood it wrong).
- Why we have selected this particular set of Roles
- Why we have decided to only increment by one with boolean operators

juanjux: - There is only one definition, McCabe, based on cyclic graphs which maybe we could implement if one day we implement an UAST->cyclic graph converter. This definition speaks about some things like shortcut boolean tests but not about all else.
- Then there are a lot of different implementations. I speak about this in the comment with three linked examples. I choose these tokens because if you check many implementations you'll see that is the most common one.
- I'll add an explanation about the counting of boolean operator in the current implementation.

uast/utils.go
smola: Using PMD as reference would be good, since we're also doing that for NPath.

juanjux: I've updated the PR to include PMD as our reference and removed switchDefault that they don't consider.

PR #105, opened 2017-05-19
POC: Add String method to Rule using XPath like notation.

alcortesm:
This is a POC for https://github.com/src-d/backlog/issues/701, I didn't pay any attention to efficiency or style.

The description of a rule is implemented by traversing the tree of rules, generating a line of text for every node with an action: Nodes are described using XPath abbreviated notation, then an arrow and the list of actions.

There are some good examples of what text would be printed for each Rule in the folder_test.go file.

The XPath syntax is longer than we anticipated because our trees are based on attributes, not node types as in XML.  This means the whole AST-XPath translation looks a bit forced and looks awkward, but as you can see in this patch, is doable and quite efficient (efficiency can be greatly improved, see my comments in the code).

Tell me what you think and let's see if we should continue along this road or not.


uast/ann/ann.go
```diff
@@ -14,9 +17,43 @@ const (
...
+// boolean output.  The String function returns  a description of the
+// predicate in a syntax similar to XPath.
+type Predicate interface {
+	fmt.Stringer
+	Eval(n *uast.Node) bool
+}
+
+// predicate implements Predicate by using desc as the XPath
+// representation of the predicate.
+type predicate struct {
```
abeaumont: I wonder if the XPath representation could be decoupled from the predicate and action interfaces/implementation. It looks like this repr is ortoghonal to both predicates and actions (i. e. we could have n different representations and they shouldn't affect how actions and predicates are implemented.)

alcortesm: I don't think I understand your comment, the purpose of the Predicate and the Action interface is exactly that (decoupling the predicate and actions from the representation).

Can you elaborate on your point?

abeaumont: I don't understand why Predicate and Action interfaces and related types include the XPath representation (stringer interface / desc field), instead of handling this XPath representation independently. Handling it in a decoupled way wouldn't require changes to existing predicates/actions and would make adding representations easier.

That being said, I don't see an easy way to implement it in a decoupled way, considering that rules are constructed through function calls, so forget about it.

alcortesm: Thanks, I see your point now.

uast/ann/ann.go
```diff
@@ -14,9 +17,43 @@ const (
...
+	eval func(n *uast.Node) bool
+	desc string // change this to a bytes.Buffer to help gc.
```
smola: You can remove this comment: `// change this to a bytes.Buffer to help gc.`, this is the kind of comment that will stay forever there ;-)

uast/ann/ann.go
```diff
@@ -134,28 +181,33 @@ func (r *Rule) Do(actions ...Action) *Rule {
...
-
-		return n.InternalType == it
+	return &predicate{
+		eval: func(n *uast.Node) bool {
+			if n == nil {
+				return false
+			}
+			return n.InternalType == it
+		},
+		desc: fmt.Sprintf("@InternalType='%s'", it),
```
smola: I also expected this to be decoupled. The type would be `hasInternalTypePredicate` instead of `predicate`, so it would be possible to test for the type from a separate function to determine the representation.

uast/ann/ann.go
alcortesm: This is supposed to be a POC, throwaway code, I was planning to delete everything once we agree how to proceed.

alcortesm:
Updated to current sdk master.

alcortesm:
@abeaumont, @smola : I've added a new commit to this PR with the suggested changes by both of you.

Let me know if I understood you both right.  The best way to review it is by opening the commit instead of the whole PR.

PS: I have chosen to use pointer receivers in all types for consistency, even when most of them are just reference types or very small structs.

alcortesm:
Rebased on top of master.

alcortesm:
After a private conversation with @smeruelo about how to show this string representation as documentation for the drivers, we have decided, for now, to add docgen package with the functionality of generating Markdown tables of the annotations.  We will later see how to integrate that into the overall documentation.

I'll implement this right away.

alcortesm:
Here is a simple final version where `Rule.String()` outputs markdown tables.  It is ugly as hell and I don't recommend using it.

I'm working on an alternative version, where `Rule`s don't know anything about markdown, instead, they export their fields and a new package `docgen.markdown` inspect them to generate the markdown tables.

uast/ann/folder.go
```diff
@@ -0,0 +1,95 @@
...
+	"bytes"
+	"fmt"
+	"strings"
+)
+
+// folder as in "analyzer of recursive data structures combining the
+// information in its nodes", not as in "directory".  This particular
+// folder traverses Rules in pre-order, generating their string
+// XPath-like descriptions.
+type folder struct {
```
smola: I would imagine `folder` as an interface implementing the `Fold()` method. Since this is an implementation, it should have a name descriptive of what it is actually doing.

alcortesm: Excellent point, done.

uast/ann/ann.go
```diff
@@ -53,14 +53,20 @@ type Rule struct {
 	rules      []*Rule
 }
 
+const (
```
abeaumont: I wouldn't make this code public, since it's only needed in the `String()` method. I guess you're making it public to reuse it in the tests, which I think it's a bad policy, but feel free to ignore this comment if the opposite is agreed on.

alcortesm: What public code are you referring to? (your comment points to line 56 which hardly has any code except for the  "const (").

If you are talking about the head constant in line 57, it is not public, it is private (its first letter is lowercase).  Also, the tests are in the same package (`ann`) as the rest of the code in this patch, meaning they can access private elements without the need of making them public.

abeaumont: Yes, I was refering to the `head` constant, which I thought was public. You're absolutely right.

alcortesm: :+1:

alcortesm:
As this is no longer a POC, I'm closing this for #125.

PR #108, opened 2017-05-22
add driver.Client to communicate with drivers

smola:
Added a convenience interface to communicate with bblfsh drivers directly. This will be used by bblfsh/server but might be used separately.


smola:
Depends on PR https://github.com/bblfsh/sdk/pull/107

PR #110, opened 2017-05-23
driver: do not return error on ParseUAST

smola:
Now driver.UASTParser conforms to protocol.Parser. That is,
it does not use a second return value for error and uses the
error handling of ParseUASTRequest.


smola:
Depends on https://github.com/bblfsh/sdk/pull/109

PR #111, opened 2017-05-25
UAST: Added role for expressions interpolated in strings

juanjux:
StringInterpolatedExpression represents a symbol or expression inserted in
some way in the string that will be evaluated and replaced by its result. This
is similar to the $value or $(value) of many languages or the {expression} of Python's fstrings.


juanjux:

PR #112, opened 2017-05-25
UAST: Added Annotation role

juanjux:
// Annotation represents the concept of some languages of symbols that
	// doesn't have a runtime effect but serve to provide information to
	// the compiler or external tools like linters (e.g. C++/D pragma, Python type annotations,
	// Java @annotations).


juanjux:

PR #113, opened 2017-05-25
UAST: Added Async and Await roles

juanjux:
// Async is a role that usually will be combined to others to indicate
	// asynchronous operation, e.g. with FunctionDeclaration declares a function
	// that will always run asynchonously, with Call declares that only a specific
	// call is to be run asynchnously (like Go's "go f()", with ForEach could mark
	// a concurrent ForEach, etc.
	Async
	// Await represents a waiting for the completion of asynchronous operations
	Await


juanjux:

PR #114, opened 2017-05-25
UAST: Added global/closure capture roles

juanjux:
// SymbolCaptureExpression is for use with some languages that support specfic
	// keywords or syntax to capture globals or outer (typically for
	// closures and lambdas) variables. These depending on the language
	// can be specified for value or reference access. These several roles should be
	// combined to express the specific way the capture is done.
	// (e.g. Python's "global" and "nonlocal" keywords, C++ captures, Kotlin receivers).
	SymbolCaptureExpression
	SymbolCaptureAll
	SymbolCaptureGlobal
	SymbolCaptureClosure
	SymbolCaptureValue
	SymbolCaptureReference


juanjux:

PR #117, opened 2017-05-30
[cli] Rename bblfsh-tools to bblfhs-sdk-tools

abeaumont:


cli/bblfsh-sdk-tools/main.go
```diff
@@ -13,7 +13,7 @@ var version string
 var build string
 
 func main() {
-	parser := flags.NewNamedParser("bblfsh-sdk", flags.Default)
+	parser := flags.NewNamedParser("bblfsh-sdk-tools", flags.Default)
```
smola: grep `bblfsh-tools` on `etc` and `test` directory. Replace remaining occurrences, run `make` to update bindata and `go install ./cmd/...`; you can then validate by doing `make test` of a driver

abeaumont: Fixed

smola:
Note that building drivers after merging this PR will require `bblfsh-sdk prepare-build` (but not `bblfsh-sdk update`).

PR #118, opened 2017-05-30
protocol: add Language and Filename to ParseUASTRequest

smola:
* These are required in most cases by the server to perform language
  detection.
* Language is also useful for the user to override language detection
  as needed.


protocol/generated.proto
```diff
@@ -12,7 +12,9 @@ option go_package = "protocol";
 message ParseUASTRequest {
 	option (gogoproto.goproto_getters) = false;
 	option (gogoproto.typedecl) = false;
-	string content = 1;
+	string filename = 1;
+	string language = 2;
```
juanjux: If the user wants to override the language version (for example python 2 vs 3) will this version be embedded in the language? (like in "python2") or should we add a new "version' optional field?

smola: It should not be included in the language field, otherwise it breaks a lot of our assumptions about language name being the same in the protocol, simple-linguist and driver image names.
We might have to add a separate language version field though.

Issue #119, opened 2017-06-01
Improve driver bootstrapping


abeaumont:
The initial code block described at [babelfish documentation](https://doc.bblf.sh/driver/sdk.html#creating-a-driver-for-a-new-language) (creating a directory, a new git repo, etc.) could be done by `bblfsh-sdk init` command

smola:
Now merged.

PR #123, opened 2017-06-02
Improved driver initialization

abeaumont:
As commented in #119


abeaumont:
@juanjux they're hook samples created by a `git init` command
@smola I wanted to avoid having an external dependency (even if the user is supposed to have it). Can change it though.

smola:
Note that `git init` might do additional validations and uses relevant global git configuration of the user.

abeaumont:
Replaced git initialization files by `git init`

PR #125, opened 2017-06-08
Autogenerate conversion tables for UAST

alcortesm:
Fixes https://github.com/src-d/backlog/issues/701.

- ann: Add Rule.String to create markdown tables using XPath like notation

  Every predicate is now a type, implementing the new Predicate interface
  and they know how to print themselves in XPath-like notation.

  The processing of the rule tree is implemented as a folding operation
  by the new xPathDescription type.

- makefile: Add new docgen target

  This adds the new rule `docgen` to the makefile, that generates
  documentation for the driver.

  It also adds docgen as a prerequisite to build.

  Right now it only prints the rules as a markdown table to a ANNOTATION.md file.

### DIFF
--- a/assets/build/bindata.go
+++ b/assets/build/bindata.go
@@ -493,6 +493,7 @@ BUILD_IMAGE=$(DOCKER_BUILD_NATIVE_IMAGE) $(DOCKER_BUILD_DRIVER_IMAGE) $(DOCKER_I
 
 # golang runtime commands
 GO_CMD = go
+GO_RUN = $(GO_CMD) run
 GO_TEST = $(GO_CMD) test -v
 GO_LDFLAGS = -X main.version=$(DRIVER_VERSION) -X main.build=$(BUILD_ID)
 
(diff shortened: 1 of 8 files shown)


smola:
IMHO this is ready as soon as we confirm in which phase we want to run it (`make build` or elsewere).

alcortesm:
FYI right now, this patch runs docgen as a prerequisite of the build target.

smola:
@alcortesm So what about putting it in a dedicated rule `docgen` at the moemnt?

alcortesm:
@smola not sure what you mean; `docgen` is already a dedicated rule (https://github.com/bblfsh/sdk/pull/125/files#diff-41b162c2528d15ff74863ca5dac59424R122).`

Right now it only generates the `ANNOTATION.md` file, but in the future, it can do more.

Right now, the `build` target has `docgen` as its prerequisite.

alcortesm:
Rebased to master.

PR #127, opened 2017-06-14
Increase JSON decoder buffer size to 4Mb

juanjux:
This fix all "buffer size excedeed" problems with all the Python files I tested (all of the modules in Python 2.7 and 3.6 stdlibs, including some huge ones) but eventually a dynamic allocation alternative will be needed.


smola:
We should use a growable buffer, but let's merge this as a temporary fix.

PR #131, opened 2017-06-17
Added Encoding field, regenerated protocol files

juanjux:
This adds a new Encoding field for the gRPC and native driver protocol `Parse[U]ASTRequest`. Current allowed values are `UTF8` (the default if unspecified) and `Base64`. 

It will try to decode Base64-encoded content into UTF8 before passing it to the driver, changing the field to UTF8 if successful; if not it will leave the field and content as Base64 since we can't use non UTF8 characters with JSON. In that case it will be the native driver responsibility to decode the Base64 content.

I'll do a PR soon documenting all this. 

As semi-offtopic, I'll also add a page or section in the developer docs about testing and debugging features that cross all the components because I almost went mad with some (in retrospective stupid) problems like, not knowing that there is a server runtime cache that you've to delete for it to use the latest drivers, forgetting that some changes in the SDK will also need a new version of the driver dockers (and tagging them as latest, deleting the caches), et cetera.


dennwc:
I guess this is a part of some offline discussion, but I was wondering why this change is needed.

Currently the request will be sent via binary gRPC connection, thus additional base64 encoding is useless, because it encodes the message to protobuf anyway. And even assuming that response will be sent as JSON, the base64 encoding is also useless - `[]byte` field will be automatically converted to/from base64 by Go, so there is no need to add an Encoding parameter.

juanjux:
@dennwc this feature is for providing a way to encode non UTF-8 encodable characters along the full pipe from the library client to the native driver independently of the communication and serialization protocols used. The gRPC and Protobuf dependency is probably what we'll go on for the foreseable future but not set in stone - I recently proposed on the public Slack channel evaluating JSONRPC since it will be easier to use/implement for clients in more languages than the gRPC/Protobuf combo.

For example, the language parser communicates by JSON and stdin/stdout and not all JSON implementations convert non-UTF8 convertible chars to Base64 (for example Python's implementation doesn't) so at a minimum at that point it should be added to the go driver -> native parser protocol and while we are at it we can just add it to the external gRPC protocol too. Almost all languages have a Base64 encoder/decoder and even if they don't it's pretty trivial to implement one.

dennwc:
@juanjux Thanks for detailed explanation! Now I understand why it will make a difference.

juanjux:
Np! Its not a mandatory field anyway with the current implementation - if missing it will default to UTF8.

protocol/driver/driver.go
```diff
@@ -148,12 +148,15 @@ func (c *parseNativeASTCommand) Execute(args []string) error {
...
 	nc, err := native.ExecClient(c.NativeBin)
 	if err != nil {
-		return fmt.Errorf("error executing native client: %s", c)
+		return fmt.Errorf("error executing native client: %s", err.Error())
 	}
 
 	defer func() { _ = nc.Close() }()
 
-	resp, err := nc.ParseNativeAST(&native.ParseASTRequest{Content: string(b)})
+	resp, err := nc.ParseNativeAST(&native.ParseASTRequest{
```
abeaumont: It'd be better to separate purely stylistic/unrelated changes in its own commit/PR in the future, since these changes are unrelated to the addition of the encoding field, and it makes it easier to distinguish between them.

protocol/generated.proto
```diff
@@ -1,14 +1,3 @@
-// Copyright 2017 Sourced Technologies SL
```
abeaumont: What's the reason to remove this?
Why is this removed from some files while left in other files?

protocol/native/native.go
```diff
@@ -36,7 +38,28 @@ func ExecParser(toNoder ToNoder, path string, args ...string) (*Parser, error) {
 }
 
 func (p *Parser) ParseUAST(req *protocol.ParseUASTRequest) *protocol.ParseUASTResponse {
-	nativeResp, err := p.Client.ParseNativeAST(&ParseASTRequest{Content: req.Content})
+	// If the request encoding is Base64 try to decode it as UTF8. Leave it
+	// as Base64 if it isn't possible.
```
abeaumont: I don't understand why we try to encode as UTF8 if we have explicitely been told that the encoding is base64, what's the reason?
As I understand it, it should be left to the native parser to handle this, I don't think this should be solved in a generic way.

juanjux: From the other point of view: if we can't decode it to UTF8 without problems at that point, why not doing it before passing the code to the native driver? It's not that the UTF8 encoding is going to be alias to the driver - it's the one that the JSON format supports so it should be able to handle it.

protocol/generated.proto
juanjux: Not removed - it's just that when regenerating the files from the `.proto` ones the copyright is not added and you've to add it by hand (@smola requested a feature for Proteus for having this done automatically. 

I'll add it by hand here too.

protocol/driver/driver.go
juanjux: I don't think a change like this one merits the increased workload for authors and reviewers of a separate commit and PR. I run GoFmt on all the code before committing (except when I forget to do it, of course) and it would be annoying to hunt what stylistic changes it do to add them to another commit.

abeaumont: I don't think that moving style changes to its own commit increases the workload of the reviewers, quite the opposite. Precisely by not separating them the author is delegating the burden of such hunt to the reviewers.

protocol/native/native.go
abeaumont: Sorry, I don't understand what you mean. As I understand it, only the native parser may be able to handle the file encoding (i. e. for a binary source code), and that's ok. I don't know if you mean that this assumption is not correct or anything else.

protocol/driver/driver.go
juanjux: We agree to disagree here, then.

alcortesm:
Is there any chance that the file positions of nodes in the UAST are wrong due to server and natives using different encodings? (from the point of view of the client, I mean)

juanjux:
@alcortesm interesting, I would expect the native ASTs to give the column positions in graphemes (vs code points) and this seem to be the case for Python (I just did a quick test with a multicp char) but we can't really make assumptions about the native AST produced for every language, anyway now that I'm disabling the auto decoding the decoding from Base64 done by the native driver should hopefully produce the same encoding so I'm not really sure that we'll find this problem.

protocol/native/native.go
juanjux: It was meant to cover cases of the client encoding stuff to Base64 that could really be encoded without problems in UTF8 (think most language specific pre unicode encodings, ISO-8859-X, etc) but @alcortesm got a good point about possible changing file line positions which could bit us if we convert to UTF8 and the original encoding was different (and used code points for positions instead of graphemes for example), so I'm disabling this on my next commit.

alcortesm:
:+1:

Issue #130, opened 2017-06-16
Make JSON decoder buffer size dynamic


abeaumont:
A better solution for this kind of problems: https://github.com/bblfsh/sdk/pull/127

juanjux:
Fixed by PR #135 when merged.

bzz:
Wow! So awesome, did not realize #135 is merged - going to test now.

bzz:
I have tested the the latest `bblfsh server` \w #135 merged and I still get `buffer size exceeded` sometimes :/ not sure if that is related to buffer size.

Here is the file it happened on https://github.com/bit-man/bit-tools/blob/master/GenUnitTests-ECI2012/src/test/java/ej2/conInvariante/RandoopTestConInvariante1.java

I would be happy to just skip such files, **but** after this error - server hangs forever.

juanjux:
Very odd since that message is not in the server anymore, I'll try with the docker image.

juanjux:
I can't reproduce it. Steps I took (using the workaround posted on Slack for the problem with the current server and env variables):

```
# Remove old images
docker rm (docker ps -a -q)  
docker rmi --force (docker images -q)

# workaround
export BBLFSH_DRIVER_IMAGES="a=a"
# Run the server
docker run -e BBLFSH_DRIVER_IMAGES --privileged -p 9432:9432 --name bblfsh bblfsh/server

# Send the file
docker run  -v $(pwd):/work --link bblfsh bblfsh/server bblfsh client --address=bblfsh:9432 /work/shit.java

time="2017-07-05T19:27:06Z" level=debug msg="reading file: /work/shit.java" 
time="2017-07-05T19:27:06Z" level=debug msg="dialing server at bblfsh:9432" 
time="2017-07-05T19:27:06Z" level=debug msg="instantiating service client" 
time="2017-07-05T19:27:06Z" level=debug msg="sending request" 

(super huge UAST is printed)
```

I tried again with several consecutive requests just in case it was hanging at the second request like in server bug #36 but it works perfectly.

bzz:
Ok, I was missing `--transport=docker-daemon` when running server, so I guess drivers were fetched from DockerHub instead of using latest build/taggeg :\

But this does not explain why `buffer size exceeded` show up, does it?

juanjux:
Actually, I should be using the version from Dockerhub to test it, let me test again without that.

juanjux:
[...]

PR #135, opened 2017-06-30
Fix hangs when the reading from the native driver raised an error before all input was consumed

juanjux:
This fix was trivial to code,  easy to explain but was a hell to debug. 

When there was any error on the reader from the native driver, the driver stdin was not being consumed until the next newline. This caused that on the next request it would still trying to write to its stdout while the Go wrapper would try to read the new request into its stdin, hanging.

This also changes the read to be done by reading buffer-sized chunks into a slice so this should also fix the "buffer size exceeded" problem.

The driver images should be rebuild once this is merged since it's a pretty annoying bug.

### DIFF
--- a/protocol/jsonlines/decoder.go
+++ b/protocol/jsonlines/decoder.go
@@ -3,8 +3,8 @@ package jsonlines
 import (
 	"bufio"
 	"encoding/json"
-	"fmt"
 	"io"
+	"fmt"
 )
 
 const (
@@ -15,6 +15,7 @@ const (
 
 type lineReader interface {
 	ReadLine() ([]byte, bool, error)
+	ReadSlice(delim byte) (line[] byte, err error)
 }
 
 // Decoder decodes JSON lines.
@@ -40,17 +41,46 @@ func NewDecoder(r io.Reader) Decoder {
 	return &decoder{r: lr}
 }
 
+// Read and discard everything until the next newline
+func (d *decoder) discardPending() error {
+	for {
+		_, err := d.r.ReadSlice('\n')
+		switch(err) {
+		case io.EOF:
+			return nil
+		case bufio.ErrBufferFull:
+			continue
+		case nil:
+			continue
+		default:
+			return err
+		}
+	}
+}
+
 // Decode decodes the next line in the reader.
 // It does not check JSON for well-formedness before decoding, so in case of
 // error, the structure might be half-filled.
 func (d *decoder) Decode(v interface{}) error {
-	line, isPrefix, err := d.r.ReadLine()
-	if isPrefix {
-		return fmt.Errorf("buffer size exceeded")
-	}
+	var line []byte
 
-	if err != nil {
-		return err
+	for {
+		chunk, isPrefix, err := d.r.ReadLine()
+		if err != nil {
+			if !isPrefix {
+				if discardErr := d.discardPending(); discardErr != nil {
+					return fmt.Errorf("%s; aditionally, there was an error " +
+					                 "trying to dicard the input buffer: %s",
+					                 err, discardErr)
+				}
+			}
+			return err
+		}
+		line = append(line, chunk...)
+
+		if !isPrefix {
+			break
+		}
 	}
 
 	switch o := v.(type) {
--- a/protocol/jsonlines/decoder.go
+++ b/protocol/jsonlines/decoder.go
@@ -3,8 +3,8 @@ package jsonlines
 import (
 	"bufio"
 	"encoding/json"
-	"io"
 	"fmt"
+	"io"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


juanjux:
Fixes bblfsh/server/issues/36.
Fixes #130.

juanjux:
PS: we should really add a lot more verbose debug logging now that it's selectable. Maybe also set the default logging level to "info". It would make it a lot easier to debug problems since when testing with the server and drivers in containers and gRPC calls you can't just use gdb. Should I open an issue?

protocol/jsonlines/decoder.go
```diff
@@ -15,6 +14,7 @@ const (
 
 type lineReader interface {
 	ReadLine() ([]byte, bool, error)
+	ReadSlice(delim byte) (line[] byte, err error)
```
abeaumont: Is it possible to replace `ReadLine` with `ReadSlice` instead of having both?

protocol/jsonlines/decoder.go
```diff
@@ -40,17 +40,41 @@ func NewDecoder(r io.Reader) Decoder {
 	return &decoder{r: lr}
 }
 
+func (d *decoder) discardPending() error {
+	// Read and discard everytihng until the next newline
```
abeaumont: s/everytihng/everything/

abeaumont: This line may be better above the function declaration, to document the function

protocol/jsonlines/decoder.go
```diff
@@ -40,17 +40,41 @@ func NewDecoder(r io.Reader) Decoder {
...
+	for {
+		_, err := d.r.ReadSlice('\n')
+		if err != nil {
```
abeaumont: this nested `if` and `if-else` may get simpler with a `switch`

protocol/jsonlines/decoder.go
```diff
@@ -40,17 +40,41 @@ func NewDecoder(r io.Reader) Decoder {
...
-	if isPrefix {
-		return fmt.Errorf("buffer size exceeded")
-	}
+	var line []byte
+
+	for {
+		chunk, isPrefix, err := d.r.ReadLine()
+		if err != nil {
+			if !isPrefix {
+				d.discardPending()
```
abeaumont: return value of this call should be checked, shouldn't it?

protocol/jsonlines/decoder.go
juanjux: Mmm, I think the code is simpler this way; [ReadSlice] https://golang.org/pkg/bufio/#Reader.ReadSlice returns a pointer (or pointers since it's a slice) to the original slice and don't block so it's perfect for the read-discard done on discardPending (no copying and no blocking). ReadLine on the other hand does a copy but also returns a specific code if the buffer is full so we can continue iterating by chunks.

Both could be implemented using the other but I think it's simpler that way in the places they're used here.

protocol/jsonlines/decoder.go
juanjux: Fixed*2, thanks.

protocol/jsonlines/decoder.go
juanjux: Mmm, it's already an error handler so my logic was to return the original error, but on a second tough I'll also check that error and produce a new error with a combined message since any error to discard the input buffer could bring more misteriously hanged containers.

protocol/jsonlines/decoder.go
juanjux: Definitely.

protocol/jsonlines/decoder.go
juanjux: PS: is probably undefined behavious to do this with the bufio object once it has already failed, but I checked causing the failure and without the discard the next call would hang so it's probably the lesser evil.

protocol/jsonlines/decoder.go
abeaumont: I see, thanks for the explanation.

protocol/jsonlines/decoder.go
```diff
@@ -3,8 +3,8 @@ package jsonlines
 import (
 	"bufio"
 	"encoding/json"
-	"fmt"
 	"io"
+	"fmt"
```
smola: apply gofmt

protocol/jsonlines/decoder.go
```diff
@@ -40,17 +41,46 @@ func NewDecoder(r io.Reader) Decoder {
...
 // error, the structure might be half-filled.
 func (d *decoder) Decode(v interface{}) error {
-	line, isPrefix, err := d.r.ReadLine()
-	if isPrefix {
-		return fmt.Errorf("buffer size exceeded")
-	}
+	var line []byte
+
+	for {
+		chunk, isPrefix, err := d.r.ReadLine()
```
smola: Maybe move this code block to a smaller function (e.g. `readLine(r lineReader) error`) to avoid too much nesting and improve readability.

protocol/jsonlines/decoder.go
juanjux: Done.

protocol/jsonlines/decoder.go
juanjux: Good idea. I also added a couple comments and simplified the return logic.

Issue #126, opened 2017-06-13
Roles numeric values do not match in Go and Proto


smola:
Roles numeric values do not match in Go and Proto. It seems they start at index 1 in Go and they start at index 0 in Proto.

It *might* be caused by a proteus issue: https://github.com/src-d/proteus/issues/86
(or it might not)

Once we fix this, we need to build new versions of published drivers.

smola:
There's a workaround for the proteus stuff: https://github.com/src-d/proteus/issues/86#issuecomment-308339286

vmarkovtsev:
Anyone? Although there is definitely a working workaround, this one is very nasty.

juanjux:
@vmarkovtsev just merged this:

https://github.com/bblfsh/sdk/pull/139

abeaumont:
As noted in #139, autogenererated file update is pending.
I still cannot assign to @mcarmonaa, so I'll do the PR with the autogenerated files myself.

abeaumont:
Solved with #142. This should be fixed now and drivers should be updated.

PR #139, opened 2017-07-07
UAST: Added Invalid Role to be zero value in protobuf

mcarmonaa:


uast/uast.go
```diff
@@ -28,11 +28,11 @@ type Hash uint32
 type Role int16
 
 const (
-	_ = iota
+	Invalid Role = iota
```
juanjux: Maybe a little comment like:

```go
// Protobuf enums start with 0 so we assign it the Invalid Role
```

or something like that so whoever see the code understand the need for this.

juanjux:
Fixes #126

abeaumont:
Sorry, I somehow missed this PR. Still have some comments though:
* I don't think this solves the issue, since the autogenerated files have not been regenerated
* The issue #126 is assigned to me, and I've been working on this. I've it mostly solved, but have been unable to create a PR because I've not been able to test it yet. We should prevent duplicating work this way.

PR #142, opened 2017-07-12
Fix role number consistency

abeaumont:


juanjux:
@abeaumont the travis build breaks with:

`protocol/generated.pb.go:599: undefined: github_com_bblfsh_sdk_uast in github_com_bblfsh_sdk_uast.Node`

abeaumont:
I see. That change is unrelated, but included it just because it changed. I guess this is something that the generator doesn't handle  properly and it's fixed by hand... removed.

abeaumont:
Fixes #126

PR #161, opened 2017-08-10
New feature to promote string properties to nodes

juanjux:
New feature to promote string properties to full nodes, similar to the existing one that allows to promote the keys in properties holding arrays to nodes.

Like in the previous case the key name will be the `internalType` and the string value will be set as the node `Token`.

This allows to convert something like:

```python
import numpy as np
```

With native (simplified) AST:

```python
                        {
                            "ast_type": "alias",
                            "asname": "np",
                            "name": "numpy"
                        }
```

To this (simplified & prettified) UAST:
                 
```python
{
  .  .  .  .  .  Roles: ImportPath,SimpleIdentifier
  .  .  .  .  .  TOKEN "numpy"
  .  .  .  .  .  Children: {
  .  .  .  .  .  .  0: alias.asname {
  .  .  .  .  .  .  .  Roles: ImportAlias,SimpleIdentifier
  .  .  .  .  .  .  .  TOKEN "np"
  .  .  .  .  .  .  .  Properties: {
  .  .  .  .  .  .  .  .  promotedPropertyString: true
  .  .  .  .  .  .  .  }
  .  .  .  .  .  .  }
  .  .  .  .  .  }
}
```

### DIFF
--- a/protocol/native/objecttonoder.go
+++ b/protocol/native/objecttonoder.go
@@ -85,6 +85,15 @@ type ObjectToNoder struct {
 	// If this option is set, all properties mapped to a list will be promoted to its own node. Setting
 	// this option to true will ignore the PromotedPropertyLists settings.
 	PromoteAllPropertyLists bool
+	// PromotedPropertyStrings allows to convert some properties which value is a string
+	// in the native AST as a full node with the string value as Token like:
+	//
+	// "SomeKey": "SomeValue"
+	//
+	// that would be converted to a child node like:
+	//
+	// {"internalType": "SomeKey", "Token": "SomeValue"}
+	PromotedPropertyStrings map[string]map[string]bool
 	// TopLevelIsRootNode tells ToNode where to find the root node of
 	// the AST.  If true, the root will be its input argument. If false,
 	// the root will be the value of the only key present in its input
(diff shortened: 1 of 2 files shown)


juanjux:
PS: also fixed a unittesst that failed to check the result of the other similar feature.

abeaumont:
I see a pair of problems with this approach:
* The original ast is being modified and custom nodes are being added to the tree. I thought this was something we didn't want to do, maybe I'm wrong.
* These new nodes lack the information the rest of nodes have, like position information.

juanjux:
@abeaumont 

- It's a pragmatic thing we already needed to do in the past to get a reasonable UAST in the Python's case (probably will happen with other ASTs in the future) since it heavily uses node properties to store other elements that should be nodes, but we don't have defined usage of properties outside of positions and token. This is something that we need to iterate on future versions of the UAST specification. 

- Unfortunately we don't have position information for these properties. We could try to extract it using the tokenizer in the future, but it's maybe too much effort (considering we don't even have the line number) for something that could change in a future version of the UAST.

abeaumont:
@juanjux Thanks for the explanation. According to documentation there **should** be start position, so I guess it's ok not to have it for some valid reason.

PR #165, opened 2017-08-25
protocol: Add Version request and response to the protocol

abeaumont:


abeaumont:
Ok, I guess you should merge this now that you're back :)

PR #166, opened 2017-08-29
New ToNoder option to fill missing position elements. Unified parser.go skeleton.

juanjux:
Part of src-d/backlog/issues/889

### DIFF
--- a/etc/skeleton/driver/normalizer/parser.go
+++ b/etc/skeleton/driver/normalizer/parser.go
@@ -10,25 +10,23 @@ var ToNoder = &native.ObjectToNoder{}
 
 // ParserBuilder creates a parser that transform source code files into *uast.Node.
 func ParserBuilder(opts driver.ParserOptions) (parser driver.Parser, err error) {
-	psr, err := native.ExecParser(ToNoder, opts.NativeBin)
+	parser, err = native.ExecParser(ToNoder, opts.NativeBin)
 	if err != nil {
-		return psr, err
+		return
 	}
 
 	switch ToNoder.PositionFill {
-	case native.None:
-		parser = psr
 	case native.OffsetFromLineCol:
 		parser = &driver.TransformationParser{
-			Parser:         psr,
+			Parser:         parser,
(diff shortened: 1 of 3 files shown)


etc/skeleton/driver/normalizer/parser.go
```diff
@@ -5,12 +5,29 @@ import (
...
 
+// ToNoder specifies the driver options. Driver programmers should fill it
+var ToNoder = &native.ObjectToNoder{}
+
 // ParserBuilder creates a parser that transform source code files into *uast.Node.
-func ParserBuilder(opts driver.ParserOptions) (driver.Parser, error) {
-	toNoder := &native.ObjectToNoder{}
-	parser, err := native.ExecParser(toNoder, opts.NativeBin)
+func ParserBuilder(opts driver.ParserOptions) (parser driver.Parser, err error) {
+	psr, err := native.ExecParser(ToNoder, opts.NativeBin)
```
abeaumont: There's no need for a different (and harder to understand) `psr` name, is it? You can just reassign `parser` if needed.

etc/skeleton/driver/normalizer/parser.go
```diff
@@ -5,12 +5,29 @@ import (
...
 	if err != nil {
-		return nil, err
+		return psr, err
```
abeaumont: If using named return values, and `parser` is used instead of `psr`, `return` would be enough here.

etc/skeleton/driver/normalizer/parser.go
```diff
@@ -5,12 +5,29 @@ import (
...
+	}
+
+	switch ToNoder.PositionFill {
+	case native.None:
+		parser = psr
```
abeaumont: No need for this case if `psr` is named `parser`

etc/skeleton/driver/normalizer/parser.go
juanjux: I wanted to differentiate the normal parser from the transformation parsers that take the first as one of their fields (2 and 3 cases in the switch), but maybe using a single parser var is clearer.

PR #175, opened 2017-09-21
Replace ReadLine for ReadBytes

juanjux:
Replaces `bufio.Reader.ReadLine` for `ReadBytes('\n')` which simplifies the code (no need to read by chunks or discard pending buffer contents on error; it does both things automatically) and also avoids a hang detected in certain instances of the driver crashing.

Part of the fixes for bblfsh/server/issues/101.


juanjux:
Yes, this `ReadBytes` is even better (simpler) than `ReadSlice`. I'm taking a look at the test that is failing.

PR #176, opened 2017-09-22
Regenerate pb files

juanjux:


protocol/generated.pb.go
```diff
@@ -1,14 +1,3 @@
-// Copyright 2017 Sourced Technologies SL
```
abeaumont: Are we removing these comments then? If so, README should be updated.

juanjux: Yes, that's what I understood from the slack chat. README updated.

PR #178, opened 2017-09-24
mayor driver sdk refactor

mcuadros:
Sorry for how big is this PR, but was more or less a non-stop working weekend. 

Improvements:
- Communication between server and driver based on grpc, the drivers now use the `protocol` package and they are able to server a grpc server via tcp, for testing and unix sockets for server (configurable via cli)
- Package `driver` and other move outside of `protocol` since is not part of the protocol, now they are under `sdk` package.
- Manifest is read and can be access by the `driver` package. 
- Protocol: New NativeParseRequest, to be able to retrieve original AST, via gRPC
- Protocol: `VersionResponse` now it based on manifest.
- Integration testing: now we don't use anymore a bash, its based on `go test` and the test are perform via `bblfsh-sdk-tools test` or `make integration-test`, the test runs against any gRPC server implementing the protocol.


sdk/driver/native.go
```diff
@@ -0,0 +1,159 @@
...
+	"io"
+	"os"
+	"os/exec"
+	"strings"
+	"sync"
+
+	"gopkg.in/bblfsh/sdk.v1/protocol"
+	"gopkg.in/bblfsh/sdk.v1/sdk/jsonlines"
+
+	"srcd.works/go-errors.v0"
```
juanjux: gopkg.in/src-d/go-errors.v1?

sdk/server/server.go
```diff
@@ -0,0 +1,63 @@
+package server
+
+import (
+	"net"
+
+	"gopkg.in/bblfsh/sdk.v1/protocol"
+
+	"github.com/Sirupsen/logrus"
+	"google.golang.org/grpc"
+	"srcd.works/go-errors.v0"
```
juanjux: gopkg.in/src-d/go-errors.v1?

etc/build/etc/it.bash
```diff
@@ -1,96 +0,0 @@
-#!/bin/bash -e
```
juanjux: :+1:

protocol/service_test.go
```diff
@@ -191,3 +177,39 @@ func checkError(err error) {
 		panic(err)
 	}
 }
+
+// ServiceMock implements protocol.Service interface implement the methods to
```
juanjux: Maybe better: "ServiceMock implements the protocol.Servce interface and the methods to be used in the test"

cli/bblfsh-sdk-tools/cmd/manifest.go
```diff
@@ -1,44 +1,27 @@
...
 
 	"gopkg.in/bblfsh/sdk.v1/cli"
 	"gopkg.in/bblfsh/sdk.v1/manifest"
 )
 
-const ManifestCommandDescription = "" +
-	"prints the manifest as a list of variables ready to be evaluated by bash or make"
-
-type ManifestCommand struct {
+type manifestCommand struct {
```
abeaumont: Manifest command is not a command anymore, just some helper code to read a manifest file, shouldn't it be named accordingly?

sdk/driver/driver.go
```diff
@@ -0,0 +1,147 @@
...
+		r.Elapsed = time.Since(start)
+	}()
+
+	var ast interface{}
+	r.Response, ast = d.doParse(req.Language, req.Content, req.Encoding)
+	if r.Status == protocol.Fatal {
+		return r
+	}
+
+	var err error
```
juanjux: Is this needed? Shouldn't be enough with `r.UAST, err :=` below? (same for the `var ast interface{}` above and below).

cli/bblfsh-sdk-tools/cmd/manifest.go
mcuadros: This pattern is already present at https://github.com/bblfsh/sdk/blob/master/cli/common.go. I can change both if you want, for something like `Helper`

abeaumont: As long as we're consistent, I'm fine with it. Whatever @juanjux prefers, I guess.

sdk/driver/driver_test.go
```diff
@@ -0,0 +1,136 @@
...
+	require.NotNil(d)
+
+	err = d.Start()
+	require.NoError(err)
+
+	r := d.Parse(&protocol.ParseRequest{
+		Language: "fixture",
+		Content:  "foo",
+	})
+
```
juanjux: Nitpicking: Inconsistent newline (other tests have the NoError/NotNil test just below the calls/struct literals).

sdk/driver/driver.go
mcuadros: nope, because r.UAST is already defined.

PR #185, opened 2017-09-28
server: logger configurations

mcuadros:
In order to improve the logging in server and clients, I created a `LoggerFactory` and `Logger` interface to use exactly the same configuration in the drivers and server.

With the goal of miss the outputs of all the driver and server, I added too the "-log-fields", to pass information to the driver to log togetther the common line "

```
 $ python driver --log-level debug --log-format text --log-fields '{"language":"python", "container":"AABBCCDD"}'
[2017-09-28T13:48:13Z]  INFO python-driver version: dev-0b6ddb4-dirty (build: 2017-09-28T13:45:46Z) container=AABBCCDD language=python
[2017-09-28T13:48:13Z] DEBUG executing native binary ... container=AABBCCDD language=python
[2017-09-28T13:48:13Z]  INFO server listening in localhost:9432 (tcp) container=AABBCCDD language=python
[2017-09-28T13:48:13Z] DEBUG registering grpc service container=AABBCCDD language=python
```
Also I added the option `--log-format` to force when or when not we want use json format.

### DIFF
--- a/sdk/server/logger.go
+++ b/sdk/server/logger.go
@@ -1,0 +1,77 @@
+package server
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+	"github.com/x-cray/logrus-prefixed-formatter"
+)
+
+// Logger represents a generic logger, based on logrus.Logger
+type Logger interface {
+	Debugf(format string, args ...interface{})
+	Errorf(format string, args ...interface{})
+	Fatalf(format string, args ...interface{})
+	Infof(format string, args ...interface{})
+	Warningf(format string, args ...interface{})
+}
+
+// LoggerFactory is a helper for create logrus.Logger's
+type LoggerFactory struct {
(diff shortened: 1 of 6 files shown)


sdk/server/logger.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/x-cray/logrus-prefixed-formatter"
+)
+
+// Logger represents a generic logger, based on logrus.Logger
+type Logger interface {
+	Debugf(format string, args ...interface{})
+	Errorf(format string, args ...interface{})
+	Fatalf(format string, args ...interface{})
+	Infof(format string, args ...interface{})
+	Warningf(format string, args ...interface{})
```
juanjux: Why the `f`'s at the end?

mcuadros: because it is `Printf` format, and as is written in the doc, it comes from logrus.Logger struct

PR #186, opened 2017-10-02
protocol: *Request include timeout for waiting for a driver at bblfshd

mcuadros:
Currently at the bblfshd, all the requests has a timeout for waiting for a driver, instead of being something hardcoded or configured in the server, I believe is better having this as a per request configuration.

The comment doesn't explain this, because is a detail of implementation and in the future may change, but a timeout maybe will be still required.

### DIFF
--- a/protocol/types.go
+++ b/protocol/types.go
@@ -75,15 +75,17 @@ type ParseRequest struct {
 	// Filename is the name of the file containing the source code. Used for
 	// language detection. Only filename is required, path might be used but
 	// ignored. This is optional.
-	Filename string
+	Filename string `json:"filename"`
 	// Language. If specified, it will override language detection. This is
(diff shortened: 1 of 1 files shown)


abeaumont:
I don't think server side configuration should be handled by the client.

juanjux:
I think it's good to allow the client to configure how much it can wait for a request (just like TCP clients can set a timeout on the connection).

mcuadros:
Its configuration, is how many time you want to wait. Depending fo the use case, you can wait more or less, or maybe retry the request.

abeaumont:
As commented in person, the server should have its own timeout that can override client's one, to prevent problems like deadlock, DoS attacks, etc.

juanjux:
Yes, that too.

PR #192, opened 2017-10-10
roles: Add some new roles

abeaumont:
Signed-off-by: Alfredo Beaumont <alfredo.beaumont@gmail.com>

Some new roles to be able to have more complete annotations.
Not an exhaustive list, but will be enough to reduce the number of _Incomplete_ roles substantially.


uast/uast.go
```diff
@@ -410,6 +410,24 @@ const (
...
+	Enumeration
+
+	// Arithmetic is a type of operation
+	Arithmetic
+
+	// Relational is a type of operation
+	Relational
+
+	// Variable is a symbolic name associatend with a value
+	Variable
```
juanjux: This would be the strict definition of variables? (e. g. only with values that can change) or for the more relaxed usage of any symbol associated with data  (e. g. including immutable/const values)? If its the first case, we could also add a "Constant" role.

abeaumont: The latter (matching the docstring). I agree a constant role may make sense, I have only added those I've used to add annotations in the java driver so far, which is not an exhaustive set, so it will probably be extended shortly

PR #196, opened 2017-10-25
Add fixtures command to bblfsh-tools to regenerate .uast and .native files

### LINKED ISSUE #195: Add a tool to bblfsh-tools to regenerate integration test fixtures

juanjux:
Fixes #195.

Example usage:

`bblfsh-sdk-tool fixtures fixtures/*.py.source`

or if the source files doesn't have an extension before the `.source` (like in the Java driver):

`bblfsh-sdk-tool fixtures --language=java fixtures/*.source`

```bash
❯ bblfsh-sdk-tools fixtures                                                                                                          jue 26 oct 2017 12:20:00 CEST
the required argument `sourcefile(s) (at least 1 argument)` was not provided
Usage:
  bblfsh-sdk-tools [OPTIONS] fixtures [fixtures-OPTIONS] [sourcefile(s)...]

Help Options:
  -h, --help               Show this help message

[fixtures command options]
      -l, --language=      Language to parse
      -e, --endpoint=      Endpoint of the gRPC server to use (default: localhost:9432)
      -n, --extnative=     File extension for native files (default: native)
      -u, --extuast=       File extension for uast files (default: uast)
      -q, --quiet          Don't print any output
          --verbose        show verbose debug information
          --root=          root of the driver (default: .)

[fixtures command arguments]
  sourcefile(s):           File(s) with the source code

Build information
  commit: 
  date:
```


common.go
```diff
@@ -1,4 +1,61 @@
...
+	"gopkg.in/bblfsh/sdk.v1/protocol"
+)
+
 const NativeBin = "/opt/driver/bin/native"
 const NativeBinTest = "/opt/driver/src/build/native"
+
+
+// NativeParseResponseToString is used for pretty print a native response. Used
+// for tests and fixtures regenetation.
+func NativeParseResponseToString(res *protocol.NativeParseResponse) (string, error) {
```
abeaumont: It'd make more sense to implement the Stringer interface instead of creating an ad-hoc method name for that.

sdk/driver/integration/suite_test.go
```diff
@@ -114,8 +115,10 @@ func (s *suite) doTestNativeParse(t *testing.T, filename string) {
 	r.Nil(err)
 
 	expected := getAST(r, filename)
+	resstr, err := common.NativeParseResponseToString(res)
```
abeaumont: `resstr` doesn't sound like the best name for a variable.

sdk/driver/integration/suite_test.go
```diff
@@ -114,35 +112,10 @@ func (s *suite) doTestNativeParse(t *testing.T, filename string) {
...
-	err := json.Unmarshal([]byte(res.AST), &s.AST)
-	r.Nil(err)
-
-	buf := bytes.NewBuffer(nil)
-	e := json.NewEncoder(buf)
-	e.SetIndent("", "    ")
-	e.SetEscapeHTML(false)
-
-	err = e.Encode(s)
+	resstr, err := NativeParseResponseToString(res)
```
abeaumont: * `resstr` is probably not the best name for this variable
* `NativeParseResponseToString` should probably be just `String`, implementing the Stringer interface.

cli/bblfsh-sdk-tools/cmd/fixtures.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	return nil
+}
+
+func (c *FixturesCommand) getNative(source string, filename string) (string, error) {
+	req := &protocol.NativeParseRequest{
+		Language: c.Language,
+		Content:  source,
+		Filename: removeSourceSuffix(filename),
```
abeaumont: This will not work for `java-driver` fixtures. If we're going to use this convention, it would be nice to a) document it somewhere and b) change existing drivers' fixtures to follow this convention.

sdk/driver/integration/suite_test.go
juanjux: Why? Is a response as string.

common.go
juanjux: [Native]ParseResponse objects already have a String() method (automatically provided from gRPC I think); this function already was in the integration tests and uses a "prettier" format keeping the fields sorted and removing the Elapsed field.

sdk/driver/integration/suite_test.go
juanjux: See above.

cli/bblfsh-sdk-tools/cmd/fixtures.go
juanjux: For the Java driver you would need to add the `--language` parameter since they doesn't have any extension that enry can use. I'm documenting the new way of doing integration tests and will add this.

sdk/driver/integration/common.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	s.Status = strings.ToLower(res.Status.String())
+	return buf.String(), nil
+}
+
+// RemoveExtension removes the last extension from a file
+func RemoveExtension(filename string) string {
```
abeaumont: This may be made simpler with something like

```go
i := strings.LastIndex(filename, path.Ext(filename))
return filename[:i]
```

Does this method needed to be public?

sdk/driver/integration/suite_test.go
abeaumont: `resstr` out of context is rather hard to understand.
If String method is used then using `res.String()` in the assertion could be enough.

sdk/driver/integration/common.go
juanjux: Yes, it's used by both the integration test suite and the fixture command that are in different subpackages (it was already in the suite, so I moved it to the `common.go` file), but I'll change it to your version.

sdk/driver/integration/suite_test.go
juanjux: Well, it's just after res that is the response :) but I'll change it to just `str` that gives context enough, I think.

The assertion can't use res.String() because that code is in the integration tests suite that checks against the generated files.

abeaumont: Hmmm, I fail to see why you cannot use/implement/move a `String` method and instead a `NativeParseResponseToString` is needed.

juanjux: Because gRPC (or Proteus) adds a default implementation of String to NativeParseReponse (in `generated.pb.go` and I didn't want to change that one, so I reused this `NativeParseResponseToString` that was already defined in the integration test suite since the rewrite.

Let me check if there is some way to provide a custom implementation for `String` in that object and it doesn't break anything.

abeaumont: I see, ok.

juanjux: I made it work with `.String()`, the generated protobuf files change so I'll update dependent projects (everything seems to work but just for peace of mind...).

PR #200, opened 2017-11-03
Remove duplicated roles. Remove old file.

### LINKED ISSUE #174: Remove duplicated roles in the nodes
With the new agglutinative UAST it'll be hard to avoid some duplicated roles. They shouldn't have any effect, but they are un-classy and annoy some users. 

It should be pretty easy to "uniq" the roles of a node in the SDK.

juanjux:
- Remove duplicated roles.
- Remove old file whose contents are now in bblfsh tools.

I'll update the drivers' fixtures once this has been merged.

Fixes #174.


uast/ann/ann.go
```diff
@@ -366,13 +366,24 @@ func AddRoles(roles ...uast.Role) Action {
...
 			}
-			n.Roles = append(n.Roles, roles...)
+			appendUniqRoles(n, roles...)
 			return nil
 		},
 		desc: strings.Join(descs, ", "),
 	}
 }
 
+func appendUniqRoles(n *uast.Node, roles ...uast.Role) {
```
abeaumont: minor: I'd use `unique` instead of `uniq`

PR #205, opened 2017-11-10
uast: ObjectToNode, IsNode, Modifier and OnToNode fields

mcuadros:
New options for ObjectToNode.

This new features are required for php-driver, I made the modification with the idea, to be as much generic as possible.

### DIFF
--- a/uast/node.go
+++ b/uast/node.go
@@ -12,7 +12,7 @@ import (
 )
 
 var (
-	ErrEmptyAST             = errors.NewKind("input AST was empty")
+	ErrEmptyAST             = errors.NewKind("empty AST given")
 	ErrTwoTokensSameNode    = errors.NewKind("token was already set (%s != %s)")
 	ErrTwoTypesSameNode     = errors.NewKind("internal type was already set (%s != %s)")
 	ErrUnexpectedObject     = errors.NewKind("expected object of type %s, got: %#v")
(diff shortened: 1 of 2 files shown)


uast/node.go
```diff
@@ -180,12 +197,23 @@ func (c *ObjectToNode) ToNode(v interface{}) (*Node, error) {
 		return nil, err
 	}
 
-	return c.toNode(root)
-}
+	nodes, err := c.toNodes(root)
```
abeaumont: I don't see the point of changing from `toNode` to `toNodes`, to have a function that will always return a single node, just to check that there's only one node in the slice... I don't see how it simplifies the code, quite the opposite.

mcuadros: it returns several nodes not only one.

uast/node.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 var (
-	ErrEmptyAST             = errors.NewKind("input AST was empty")
+	ErrEmptyAST             = errors.NewKind("empty AST given")
```
juanjux: ...ok xD

uast/node.go
```diff
@@ -293,31 +317,43 @@ func (c *ObjectToNode) toNode(obj interface{}) (*Node, error) {
 
 	sort.Stable(byOffset(n.Children))
 
-	return n, nil
+	if c.IsNode != nil && !c.IsNode(m) {
+		return n.Children, nil
```
juanjux: This will have the consequence that if the node doesn't pass the IsNode() filter the children are re-parented to m's parent. I don't know if this is intended. Maybe if we don't consider something a Node we don't want its children either?. Anyway this behaviour should be documented in IsNode.

uast/node.go
```diff
@@ -293,31 +317,43 @@ func (c *ObjectToNode) toNode(obj interface{}) (*Node, error) {
...
+	}
+
+	return []*Node{n}, nil
 }
+func (c *ObjectToNode) applyModifier(m map[string]interface{}) error {
```
juanjux: Useless function. Just do this on line 242.

Issue #203, opened 2017-11-07
Return the language when using "autodetect" feature


dpordomingo:
When it is used "autodetect" feature, passing the `filename` and `content` in the request, it would be great to obtain the `language` as part of the response.

It could be something like:
```protobuf
message ParseResponse {
	option (gogoproto.goproto_getters) = false;
	option (gogoproto.goproto_stringer) = false;
	option (gogoproto.typedecl) = false;
	gopkg.in.bblfsh.sdk.v1.protocol.Status status = 1;
	repeated string errors = 2;
	google.protobuf.Duration elapsed = 3 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
	gopkg.in.bblfsh.sdk.v1.uast.Node uast = 4 [(gogoproto.customname) = "UAST"];
	string language = 5;
}
```

instead of the current [ParseResponse](https://github.com/bblfsh/sdk/blob/master/protocol/generated.proto#L47):
```protobuf
message ParseResponse {
	option (gogoproto.goproto_getters) = false;
	option (gogoproto.goproto_stringer) = false;
	option (gogoproto.typedecl) = false;
	gopkg.in.bblfsh.sdk.v1.protocol.Status status = 1;
	repeated string errors = 2;
	google.protobuf.Duration elapsed = 3 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
	gopkg.in.bblfsh.sdk.v1.uast.Node uast = 4 [(gogoproto.customname) = "UAST"];
}
```

juanjux:
Yes. Also for `NativeParseResponse`.

juanjux:
Fixed by SDK v1.8.0 (already incorporated by latest Python and Java drivers) and bblfshd v2.2.0.

PR #208, opened 2017-11-13
Add language to Response and NativeResponse.

juanjux:
Adds the Language field to the responses. Part of #203, combined with server PR   that actually fills the language field.


mcuadros:
this change will force to update all the fixtures, since the output of the request will be different, right?

juanjux:
@mcuadros I don't think so, since the String() method of the objects, that the fixtures use, doesn't print the language, but I'll update them if needed.

juanjux:
Trough now that you said it... I should update those methods so the language is printed too since is part of the request (and the fixtures). I'll update the PR.

mcuadros:
Then lets remove to the ".source" from the fixtures

juanjux:
@mcuadros that was an AMBUSH.

I've updated this PR to support integration tests without the `.source` and made another PR for the Python driver (that won't work until this is merged, just like the bblfshd PR): https://github.com/bblfsh/python-driver/pull/136

Tomorrow I'll update the Java driver with the renamed fixtures.

Issue #214, opened 2017-11-28
sdk formats objects from AST weirdly


smacker:
AST -> UAST
```
{"langSpecific": {"a": 1}} -> {"Properties": {"langSpecific": "map[a:1]"}}
```

Such format isn't very useful.

more info here: https://github.com/bblfsh/sdk/pull/213#issuecomment-347505198

juanjux:
I see two cases here:

- The property is semantically unimportant. Then it doesn't matter what format it really have, trough mapping objects and arrays to the JSON format would be nice.

- The property is semantically important. Then it should be converted to a UAST node (or list of nodes) with roles and such using the promote* config or one of the ToNoder callbacks.

smacker:
Thank you for your opinion!

Then I see 2 possible improvements here:

- Reflect need of using promote* or modifier (in case of array) for semantically important properties here: https://doc.bblf.sh/driver/annotations.html 
- Format as JSON to make it nicer.

I'll send PR for the latter later.

juanjux:
Fixed by #215.

PR #215, opened 2017-11-28
Json values in props

smacker:
Small improvement as discussed in https://github.com/bblfsh/sdk/issues/214


juanjux:
Tested and working, despite reflect fame of being slow this doesn't seem to have much effect on performance compared to the previous conversion to string. Thanks for the PR!

PR #217, opened 2017-11-29
avoid installation when Go is already installed

campoy:


mcuadros:
How could be already installed? 

Also as you can read inside of the Dockerfile is copied from:
https://github.com/docker-library/golang/blob/master/1.8/alpine/Dockerfile

And we need to update it to use 1.9, so I don't know if this PR will be used to much.

campoy:
_note_: the link points to a 404

We have `FROM ${DOCKER_BUILD_NATIVE_IMAGE}`, and in my case Go is installed in the native image.
There's a deeper conversation about whether it'd be possible to use multistage dockerfiles soon to avoid this kind of conflicts.

smola:
@dennwc Maybe this is not relevant anymore with the new build system?

dennwc:
Yes, I think so. Closing.

PR #218, opened 2017-11-30
Avoid unnecessary panic by ignoring nil values in maps

campoy:


uast/node.go
```diff
@@ -305,6 +306,8 @@ func (c *ObjectToNode) toNodes(obj interface{}) ([]*Node, error) {
 			}
 
 			n.Children = append(n.Children, children...)
+		case nil:
```
mcuadros: is not covered by any test.

etc/build/etc/build/Dockerfile.build.alpine.tpl
```diff
@@ -11,20 +11,28 @@ RUN set -ex \
...
-	&& rm golang.tar.gz \
-	&& cd /usr/local/go/src \
-	&& patch -p2 -i /no-pic.patch \
-	&& ./make.bash \
-	\
-	&& rm -rf /*.patch \
-	&& apk del .build-deps
+		openssl
+
+ENV RUNTIME_GO_VERSION ${RUNTIME_GO_VERSION}
```
mcuadros: this code belongs to another PR

campoy: ouch, E_TOO_MANY_BRANCHES
I'll fix it

campoy:
Test added and it made me consider what to do with nil maps, which I think should also be ignored.

uast/node.go
campoy: done

PR #219, opened 2017-11-30
[Node.go] Export nil property values in the native AST as json null

juanjux:
Avoid a crash when converting null values in the native AST.


juanjux:
codecov was actually useful reminding me that I should add a test for this. Will update the PR.

PR #221, opened 2017-12-07
improve message for inorder traversal of non binary nodes

campoy:
The test would otherwise panic without explaining why.


uast/ann/ann_test.go
```diff
@@ -508,3 +508,27 @@ func TestRuleOnRulesActionError(t *testing.T) {
 	offendingNode := extraInfoError.Node()
 	require.Equal(offendingNode.InternalType, "foo")
 }
+
+func TestBetterErrorMessageForInorderTraversalOfNonBinaryNode(t *testing.T) {
```
juanjux: You won the prize for the longest function name haha.

PR #222, opened 2017-12-07
simplify iteration and avoid making copies

campoy:
Currently, for infix and postfix traversals, any modification to the parent is lost since it's performed on a copy.

This fixes the bug and simplifies the code keeping all iteration logic in a single function.


juanjux:
I see the solution if better using a map (I used the same solution for the libuast iterators) but I don't understand the bug part, where are the nodes in the stack modified during the traversal?

campoy:
Look at the test, the role `uast.Key` disappears currently when using `infix` or `postfix`

juanjux:
I see. It seems it's not working, the traversal tests are failing (wrong order and duplicates).

dennwc:
This is definitely broken on v2. Also, most of the traversal code is actually in the go-client now. Closing this for now, but feel free to apply to the go-client.

PR #227, opened 2018-01-25
support JSON encoding and decoding of Role

campoy:


campoy:
The bblfsh dashboard returns a JSON object that can't be parsed as-is unless we do this.

campoy:
The `jsonenums` binary is from https://github.com/campoy/jsonenums

I wonder if we should call `stringer` and `jsonenums` from the `.travis.yml`

juanjux:
- You should update the README to include the jsonenums dependency.
- Won't this break code that expect the roles to be numeric and use helper functions like [this one](https://github.com/bblfsh/client-python/blob/master/bblfsh/__init__.py#L26) to translate them?

PR #228, opened 2018-02-07
move common main function from template to driver package

dennwc:


dennwc:
Looks like the user profile is gone as well: https://github.com/jteeuwen

dennwc:
@juanjux Is there a script to regenerate all drivers?

juanjux:
I'll search for some fork of go-bindata and fork it inside bblfsh.

No script to regenerate existing drivers since they could have changes but we could do it as we update them for other reasons.

juanjux:
@dennwc I've forked go-bindata to https://github.com/bblfsh/go-bindata could you update the PR to link it?

juanjux:
@dennwc looks like somebody re-registered the account keeping the project so the build should work now without changes, I'll restart it.

juanjux:
@dennwc now it says that the generated bindata is out of sync, can you regenerate it and update the PR?

PR #234, opened 2018-02-20
Move driver discovery from docs repository to SDK

dennwc:
Driver discovery uses public Github API to list all supported drivers and their manifests.

This functionality will be used by [language table generator](https://github.com/bblfsh/documentation/pull/141) for docs repository and [bblfshd](https://github.com/bblfsh/bblfshd/pull/148) to get the list of official/recommended drivers.  

### DIFF
--- a/manifest/discovery/discovery.go
+++ b/manifest/discovery/discovery.go
@@ -1,0 +1,289 @@
+// discovery package implements helpers for clients to discover language drivers supported by Babelfish.
+package discovery
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"net/http"
+	"regexp"
+	"sort"
+	"strings"
+	"sync"
+
+	"github.com/google/go-github/github"
+	"gopkg.in/bblfsh/sdk.v1/manifest"
+)
+
+const (
+	GithubOrg = "bblfsh"
+)
+
+// topics that each driver repository on Github should be annotated with
+var topics = []string{
+	"babelfish", "driver",
+}
+
+// Driver is an object describing language driver and it's repository-related information.
+type Driver struct {
+	manifest.Manifest
+	repo        *github.Repository
+	Maintainers []Maintainer `json:",omitempty"`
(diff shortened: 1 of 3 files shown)


dennwc:
Note that currently `ocaml` and `rust` are listed as `alpha`, thus they will appear in the list of official drivers.

If they are not up-to-date, we should demote them to `pre-alpha` or `inactive` manually.

See https://github.com/bblfsh/sdk/issues/235 for an alternative solution.

manifest/discovery/discovery.go
```diff
@@ -0,0 +1,289 @@
...
+		m := Maintainer{Name: sub[1], Email: sub[2]}
+		if len(sub) >= 5 {
+			m.Github = sub[4]
+		}
+		out = append(out, m)
+	}
+	d.Maintainers = out
+	return nil
+}
+
```
juanjux: Should have a comment here, and above the options to have the godoc correctly generated.

manifest/discovery/discovery.go
```diff
@@ -0,0 +1,289 @@
...
+		opt = &Options{}
+	}
+	if opt.Organization == "" {
+		opt.Organization = GithubOrg
+	}
+	cli := github.NewClient(nil)
+
+	var out []Driver
+	// list all repositories in organization
+	for page := 1; ; page++ {
```
juanjux: Can move this loop/functionality (listing all the repositories and getting the drivers). to another function.

juanjux:
I would like to have more tests for some of the new functions.

manifest/discovery/discovery.go
dennwc: Options fields are OK as-is (godoc supports both inline and header comments), but thanks for spotting missing docs for Options itself.

manifest/discovery/discovery.go
dennwc: Done

PR #239, opened 2018-03-01
Add a tool to visualize native AST

dennwc:
Typical usage:
`bblfsh-sdk-tools ast2gv --type=type --out=svg ./fixtures/*.native`

Output:
u2_type_public go native

### DIFF
--- a/cli/bblfsh-sdk-tools/cmd/ast2gv.go
+++ b/cli/bblfsh-sdk-tools/cmd/ast2gv.go
@@ -1,0 +1,182 @@
+package cmd
+
+import (
+	"bytes"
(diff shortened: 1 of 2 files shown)


juanjux:
I like it (did [something similar](https://github.com/bblfsh/python-pydetector/blob/25b6eae95abc49cf05165ad45745a6db38f53190/pydetector/astexport.py#L24) a long time ago for debugging the drivers when we were starting with bblfsh), probably there are graphviz libraries for Go to avoid generating the string and check changes in shapes, graph types and options easier if we want to do that in the future, but for such a small module that's probably overkill.

Issue #235, opened 2018-02-20
Outdated drivers are treated as up-to-date on SDK updates


dennwc:
Currently drivers manifests has no mentions of version of SDK they were written for. It leads to a situation when driver might become outdated and incompatible, while still being listed as `beta` or even `stable`.

I propose to include new (required) field to manifest with semantic version of SDK. This field will be overwritten by `bblfsh-sdk update` and will be considered when interpreting driver status.

As always, minor version difference will not affect driver status, but major difference will automatically drop the status to `inactive`.

juanjux:
We're planning to add a dependency to the driver version using glide, but this could also be additionally done.

dennwc:
Glide (why not `dep`?) will fix an issue from driver's perspective (SDK changes), but will not directly solve issue from bblfshd perspective (drivers decomes outdated).

Potentially we could read deps file and check what version is mentioned there - it will be the version of SDK drivers supports, and bblfshd can filter these values depending on SDK version it was compiled for.

juanjux:
Outside of the convenience of using dep, I think the SDK dependency should be stated in the manifest just like we do with the required Go version.

dennwc:
Is it just for quick access? Driver discovery can pull dep file as it pulls `MAINTAINERS` currently.

juanjux:
It's because the manifest should have all the important versions needed in a single place, I think (currently it has the runtime version of the language and the Go versions). Then the bblfsh-sdk update command could update the Gopkg.toml file from the manifest as it does with other things.

dennwc:
I agree that manifest should contain this information as well. I propose to parse both `MAINTAINERS` and `Gopkg.toml` as a part of `bblfsh-sdk update` and update manifest accordingly. This will allow dropping the parsing code from driver discovery and will remove some unnecessary calls to Github API.

dennwc:
The first step is done - new drivers will include an SDK version in the `Gopkg.toml`.

bzz:
🎉 What would be the next action items here?

dennwc:
First of all, the `bblfsh-sdk` tool will need to detect `vendor` directory and run the `bblfsh-sdk` from the vendored SDK. This way we can make sure that `bblfsh-sdk` is the same version as the SDK package listed in the manifest. But this can be solved separately (see https://github.com/bblfsh/sdk/issues/325).

[...]

PR #251, opened 2018-03-30
Transformation DSL (BIP-4 and BIP-5)

### LINKED ISSUE #240: Refactor internal SDK types to use schema-less representation (BIP-4)
We will move current `Node` structure to protocol and will use maps and slices to represent AST nodes internally. Conversion to `Node` will happen in protocol package to preserve backward compatibility.

All the work happens in [bblfsh/sdk@transform-dsl](https://github.com/bblfsh/sdk/tree/transform-dsl).

### LINKED ISSUE #241: Implement transformation DSL (BIP-5)
We should implement shape-based transformation DSL as described in BIP-5.

All the work happens in [bblfsh/sdk@transform-dsl](https://github.com/bblfsh/sdk/tree/transform-dsl).

**TODO:**
- [x] Transformer
- [x] Is
- [x] Var
- [x] Any
- [x] And
- [x] Not
- [x] Check
- [x] Object
- [x] Out
- [x] Arr
- [x] OtherKeys
- [x] Lookup
- [x] Helpers

### LINKED ISSUE #242: Split ToNode into separate transformations (BIP-5)
Use new DSL to port old `ObjectToNode` transformations.

Requires: #241 

**TODO:**
- [x] InternalTypeKey
- [x] OffsetKey, LineKey, ColumnKey
- [x] EndOffsetKey, EndLineKey, EndColumnKey
- [x] TokenKeys
- [x] SpecificTokenKeys
- [x] SyntheticTokens
- [x] PromotedPropertyLists
- [x] PromoteAllPropertyLists
- [x] PromotedPropertyStrings
- [x] TopLevelIsRootNode
- [x] drop OnToNode
- [x] drop Modifier
- [x] drop IsNode

### LINKED ISSUE #244: Port annotation DSL predicates (BIP-5)
Port old annotation DSL predicates to a new transformation DSL.

Requires: #241 

**TODO:**
- [x] Self
- [x] Children(HasInternalRole)
- [x] Roles
- [x] HasInternalType
- [x] HasProperty
- [x] HasToken
- [x] And
- [x] Not
- [x] drop Children
- [x] drop Descendants, DescendantsOrSelf
- [x] drop HasInternalRole
- [x] drop HasChild
- [x] drop Or

dennwc:
Final implementation of AST transformation DSL.

Resolves #240
Resolves #241
Resolves #242
Resolves #244


uast/transformer/legacy.go
```diff
@@ -98,19 +100,15 @@ func (n ObjectToNode) transformer() Transformer {
 	)
 }
 
-var _ Transformer = ObjectToNode{}
-
-type RolesDedup struct{}
-
-func (RolesDedup) Do(root uast.Node) (uast.Node, error) {
-	nroot, ok := uast.Apply(root, func(n uast.Node) (uast.Node, bool) {
+func RolesDedup() TransformFunc {
```
juanjux: These are the kind of changes that not only probably optimize everything but also makes the code easier to understand reducing layers of complexity and programmers' cognitive load. :+1:

etc/skeleton/driver/impl/impl.go
```diff
@@ -0,0 +1,8 @@
+package impl
+
+import "gopkg.in/bblfsh/sdk.v1/sdk/driver"
+
+func init() {
+	// Can be overridden to link a native driver into a Go driver server.
```
juanjux: I would have left this for another PR since this one is fat enough and this functionality is not directly related with the background of it.

juanjux:
The code in general looks pretty good to me. The only comments is that since this new system adds a non trivial level of complexity compared with the previous one and changes almost everything it's pretty hard to see how the pieces work together even if you understand them individually; thus a good tutorial is badly needed.

juanjux:
Another nitpick is that there are several different things mixed in this PR (the drivers without js, the fixtures, etc). It should be related only to the BIP4 and 5 changes and even then it could probably be split into several PRs.

etc/skeleton/driver/impl/impl.go
dennwc: This is a pretty small change, and it's required to make native integration tests to work.

dennwc:
@juanjux I understand that this PR is quite large, but trying to split it will only delay the release with only minor benefits.

The only unrelated change might be the JSON protocol omission, but it's literally a two-line change that can be reviewed in under few minutes. It was required to be able to debug the transformation pipeline, thus I think it's still related and useful as a part of this PR.

I don't think that fixtures are unrelated - it's required to develop the drivers on top of DSL faster. Without it, you will need to rebuild containers and reinstall the drivers after each change to annotations. With it, you can compile native driver once and re-run `go test` while changing annotations, which takes less than a second.

juanjux:
Well, faster is not required :) the integration tests are there because in the past we detected some differences when getting the UAST under bblfshd (for example for empty files or problems with encoding, though that was fixed already). It should be pretty fast after the first build. There is also the change of configurable drivers implementations that probably isn't required. But it's nice to take less time, as the other features are also nice, just try to split one PR for every feature next time because the review quality goes down as the number of changes increases.

uast/transformer/ast.go
```diff
@@ -0,0 +1,189 @@
...
+	// 1) roles saved from left side (if any)
+	// 2) static roles from arguments
+	// 3) roles from conditional operation
+	b.SetFieldObj(RolesFieldOp(typ+"_roles", rop, roles...))
+	return Part("_", b)
+}
+
+// ObjectRoles creates a shape that adds additional roles to an object.
+// Should only be used in other object fields, since it does not set any type constraints.
+func ObjectRoles(vr string, roles ...role.Role) Op {
```
juanjux: The `vr` parameter is neither documented or have an informative name.

mcuadros:
Holy moly, how you expect that I can review a PR with 30 commits and that amount changes. We need to improve how we make PRs, I can't review this.

uast/transformer/ast.go
```diff
@@ -0,0 +1,189 @@
...
+		Name: uast.KeyRoles,
+		Op: If(vr+"_exists",
+			Append(NotEmpty(Var(vr)), rop),
+			rop,
+		),
+	}
+}
+
+// ASTObjectLeft construct a native AST shape for a given type name.
+func ASTObjectLeft(typ string, ast ObjectOp) ObjectOp {
```
juanjux: On both the go driver and the java one I've seen you add a `mapAST` and `mapASTCustom` wrapper that calls these. They should probably be in the SDK instead of each driver.

Also, `mapAST` almost always produce this annotation code:

```go
	mapAST("MethodInvocation", Obj{
		"expression":  OptObjectRoles("expr"),
		"name": OptObjectRoles("name"),
		"arguments": Each("args", ObjectRoles("arg")),
	}, Obj{
		"expression":  OptObjectRoles("expr", role.Call, role.Receiver),
		"name": OptObjectRoles("name", role.Call, role.Callee),
		"arguments": Each("args", ObjectRoles("arg", role.Call, role.Argument, role.Positional)),
	}, role.Expression, role.Call),
```

with the second object being exactly the same as the first one with the roles added, so a wrapper could just take the second Obj and infer the first part, which would help reduce the verbosity of the drivers.

uast/transformer/ast.go
```diff
@@ -0,0 +1,189 @@
...
+		)
+	}
+	return Fields{
+		{Name: uast.KeyType, Op: String(uast.TypeOperator)},
+		{Name: uast.KeyToken, Op: Var(vr)},
+		{Name: uast.KeyRoles, Op: opRoles},
+	}
+}
+
+func uncomment(s string) (string, error) {
```
juanjux: This should be parametrized for the comment marks instead of hardcoding `//` and `/*`.

uast/transformer/ast.go
```diff
@@ -0,0 +1,189 @@
...
+
+func comment(s string) (string, error) {
+	if strings.Contains(s, "\n") {
+		return "/*" + s + "*/", nil
+	}
+	return "//" + s, nil
+}
+
+// UncommentCLike removes // and /* */ symbols from a given string variable.
+func UncommentCLike(vr string) Op {
```
juanjux: Should not be necessary if the above is parametrized.

uast/transformer/ops.go
```diff
@@ -0,0 +1,1099 @@
...
+
+func (op opIs) Check(st *State, n uast.Node) (bool, error) {
+	v2, ok := n.(uast.Value)
+	if !ok {
+		return op.v == nil && n == nil, nil
+	}
+	return op.v == v2, nil
+}
+
+func (op opIs) Construct(st *State, n uast.Node) (uast.Node, error) {
```
juanjux: Missing doc.

uast/transformer/ops.go
```diff
@@ -0,0 +1,1099 @@
...
+	copy(arr[i:], fields)
+
+	obj.fields = arr
+	return obj
+}
+
+var _ ObjectOp = Fields{}
+
+// Fields is a helper for multiple operations on object fields with a specific execution order. See Object.
+type Fields []Field
```
juanjux: I would like an expanded explanation about when to use Obj and when to use Fields in the annotations.

uast/transformer/ast.go
dennwc: These functions exist in all drivers because I expected them to use the same approach as Go driver and mix additional roles automatically (like `Statement` or `Expression`) for every `mapAST` call depending on the type. But in practice, I inlined these roles for Java, and now doing the same for JS. So I agree, it makes sense to move it to SDK as well.

Regarding the second question, drivers also define `annotateType` that solves the problem that you described. However, it only covers simple cases where all field all pointing to a node or value. The case you describe here includes `Each` and thus it cannot be simplified using `annotateType` call.

uast/transformer/ast.go
dennwc: It would make it more verbose, but it would be useful for `Javadoc`, I agree.

Are you thinking about this kind of signature?
```golang
func Uncomment(vr string, tokens [][2]string) Op
```

dennwc:
@mcuadros I understand that it's nearly impossible to review it via Github without trying the code. It should be more manageable with tools like [reviewable.io](https://reviewable.io). In any case, once I get all feedback from @juanjux (as a form of crash-test), I will rearrange and squash commits so they are better clustered by feature.

dennwc:
> change of configurable drivers implementations that probably isn't required

@juanjux Can you elaborate, please?

uast/transformer/ast.go
juanjux: That could work, yes.

dennwc: Found that it won't work - current version reverses the transform by checking for `\n`. Making it generic won't work this way. Still, we can provide one helper to remove a single token pair.

dennwc:
@juanjux Okay, will move it there

juanjux:
Thank you for the rebasing. With the usability improvements and the documentation (in the other repo) this LGTM.

juanjux:
Remainder, so we don't forget, to recover the feature to automatically mark nodes without any roles with the "Unannotated" role. The role is not optional in that case so I don't think adding it to the drivers skeleton files would serve (because the driver programmer could remove it) but if that's the only option is better than nothing.

dennwc:
@juanjux Implemented `Unannotated` helper as well

---

Reviewed 9 of 38 files at r1, 49 of 74 files at r3, 6 of 9 files at r4, 20 of 20 files at r5, 2 of 2 files at r6.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[uast/transformer/ast.go, line 92 at r4](https://beta.reviewable.io/reviews/bblfsh/sdk/251#-L9j8G_Ip5FAsG0pOYmW-r4-92:-LAYk4jlEM0lz0FjpQ1N:b-hoxnzb) ([raw file](https://github.com/bblfsh/sdk/blob/7b4142417ece93b67a20285c2e1fa488b8ebd9b4/uast/transformer/ast.go#L92)):*

These functions exist in all drivers because I expected them to use the same approach as Go driver and mix additional roles automatically (like `Statement` or `Expression`) for every `mapAST` call depending on the type. But in practice, I inlined these roles for Java, and now doing the same for JS. So I agree, it makes sense to move it to SDK as well.

Regarding the second question, drivers also define `annotateType` that solves the problem that you described. However, it only covers simple cases where all field all pointing to a node or value. The case you describe here includes `Each` and thus it cannot be simplified using `annotateType` call.
</blockquote></details>

Added helpers, please mark as resolved.

---

*[uast/transformer/ast.go, line 187 at r4](https://beta.reviewable.io/reviews/bblfsh/sdk/251#-L9j8G_Ip5FAsG0pOYmW-r4-187:-LAYjKxS-QHvqSjjSgQz:b-9zrxo5) ([raw file](https://github.com/bblfsh/sdk/blob/7b4142417ece93b67a20285c2e1fa488b8ebd9b4/uast/transformer/ast.go#L187)):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

Should not be necessary if the above is parametrized.
</blockquote></details>

Added a helper for a single token pair. `UncommentCLike` still remains, because it uses custom logic to detect what token to use for a reverse transform.

---

*[uast/transformer/ops.go, line 47 at r4](https://beta.reviewable.io/reviews/bblfsh/sdk/251#-L9j8G_Ip5FAsG0pOYmZ-r4-47:-LAYjgZB0wzykXBFAHiV:bfuubcj) ([raw file](https://github.com/bblfsh/sdk/blob/7b4142417ece93b67a20285c2e1fa488b8ebd9b4/uast/transformer/ops.go#L47)):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

Missing doc (also in other exported symbols in this module).
</blockquote></details>

Already added docs for exported symbols. Docs for all unexported types is in their struct docs instead of methods.

---

*[uast/transformer/ops.go, line 211 at r4](https://beta.reviewable.io/reviews/bblfsh/sdk/251#-L9j8G_Ip5FAsG0pOYmZ-r4-211:-LAYjyeJAM4VkVqRF4Sp:b-sdgwhr) ([raw file](https://github.com/bblfsh/sdk/blob/7b4142417ece93b67a20285c2e1fa488b8ebd9b4/uast/transformer/ops.go#L211)):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

I would like an expanded, more practical, explanation about when to use `Obj` and when to use `Fields` in the annotations (like the need to use a `LockupArrOpVar` call).
</blockquote></details>

Will be done in documentation repository

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/bblfsh/sdk/251#-:-LAZTDkO6oIXTkWIiIz0:b-qearwz)*

juanjux:
Review status: 56 of 64 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/bblfsh/sdk/251#-LAqlYtm2BsYBsK_TDyA:-LAqlYtm2BsYBsK_TDyB:b-lv5t5g) (no related file):*
Feature request: with the previous DSL you could do:

```go
On(HasInternalRole("left")).Roles(uast.Left),
On(HasInternalRole("right")).Roles(uast.Right),
```

This was extremely convenient for properties that could always be tagged with the same roles no matter what the parent node type was. AFAIK, this can't be done with the new DSL (you always need to annotate the property nodes in the context of a parent type, correct me if I'm wrong) but it would be great to recover it. Would also simplify a lot the annotation of operators in some languages without needing helpers.

If done, doesn't have to go into this PR, but I don't want to add issues to the project until this is merged into ~master.

---


*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/251)*

dennwc:
Reviewed 8 of 8 files at r7.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://beta.reviewable.io/reviews/bblfsh/sdk/251#-LAqlYtm2BsYBsK_TDyA:-LAqmWWI2IBBs9bcCHeR:bo26jz5) (no related file):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

Feature request: with the previous DSL you could do:

```go
On(HasInternalRole("left")).Roles(uast.Left),
On(HasInternalRole("right")).Roles(uast.Right),
```

This was extremely convenient for properties that could always be tagged with the same roles no matter what the parent node type was. AFAIK, this can't be done with the new DSL (you always need to annotate the property nodes in the context of a parent type, correct me if I'm wrong) but it would be great to recover it. Would also simplify a lot the annotation of operators in some languages without needing helpers.

If done, doesn't have to go into this PR, but I don't want to add issues to the project until this is merged into ~master.
</blockquote></details>

It can be done right now, but helpers like `AnnotateTypeFields` won't work, because they expect a type to be set. Will add helpers for objects with no specific type.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/bblfsh/sdk/251)*

dennwc:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/bblfsh/sdk/251#-LAqo1asEh81mgXlxvq7:-LAqo1asEh81mgXlxvq8:b791fnm) (no related file):*
Remainder to myself: it was decided to move `Unannotated()` to the getter of `uast.Node`.

---


*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/251)*

juanjux:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/bblfsh/sdk/251#-LAqlYtm2BsYBsK_TDyA:-LAqzNpp2WGA-Vio6z4i:b-uvrt0n) (no related file):*

It can be done right now, but helpers like `AnnotateTypeFields` won't work, because they expect a type to be set. Will add helpers for objects with no specific type.
</blockquote></details>

You you post a small example about how this could be done? (without the helper).

---


*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/251)*

dennwc:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/bblfsh/sdk/251#-LAqlYtm2BsYBsK_TDyA:-LArEsMo4OnM5DqinLGp:b-owcebr) (no related file):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

You you post a small example about how this could be done? (without the helper).
</blockquote></details>

Sure, should be something like this:
```golang
Map("left-right",
   Part("m", Obj{
      "left": ObjectRoles("left"),
      "right": ObjectRoles("right"),
   }),
   Part("m", Obj{
      "left": ObjectRoles("left", role.Left),
      "right": ObjectRoles("right", role.Right),
   }),
)
```

---


*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/251)*

dennwc:
Reviewed 4 of 4 files at r8.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/bblfsh/sdk/251#-LAqlYtm2BsYBsK_TDyA:-LAs6Q2W53_3PAsOSwlY:b-aks7s4) (no related file):*

Sure, should be something like this:
```golang
Map("left-right",
   Part("m", Obj{
      "left": ObjectRoles("left"),
      "right": ObjectRoles("right"),
   }),
   Part("m", Obj{
      "left": ObjectRoles("left", role.Left),
      "right": ObjectRoles("right", role.Right),
   }),
)
```
</blockquote></details>

@juanjux So will you need a helper, or it's not a common pattern? I expected an example to be more complex, actually...

---


*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/251)*

juanjux:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/bblfsh/sdk/251#-LAqlYtm2BsYBsK_TDyA:-LAs3KX25aLSJqS_F517:b-b93por) (no related file):*

@juanjux So will you need a helper, or it's not a common pattern? I expected an example to be more complex, actually...
</blockquote></details>

Thanks! I would say it's a semi-common pattern (I added a small helper in the PHP driver) but not unavoidable (you can always tag the keys inside the types). The example works for most cases I tried but I've one that it doesn't (stmts key in the PHP driver, but it's only used in few nodes so I can annotate it there), will check with you later on slack once I've finished other annotations.

---


*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/251)*

dennwc:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/bblfsh/sdk/251#-LAqlYtm2BsYBsK_TDyA:-LAsClQu5ePoHVXFQ7Ww:b-ars5mk) (no related file):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

Thanks! I would say it's a semi-common pattern (I added a small helper in the PHP driver) but not unavoidable (you can always tag the keys inside the types). The example works for most cases I tried but I've one that it doesn't (stmts key in the PHP driver, but it's only used in few nodes so I can annotate it there), will check with you later on slack once I've finished other annotations.
</blockquote></details>

Okay, will wait for more info. Let me know if you think that we need this helper in SDK.

---


*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/251)*

juanjux:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/bblfsh/sdk/251#-LAqlYtm2BsYBsK_TDyA:-LAsI58q1J2bTpbJPNb7:beoflrx) (no related file):*

Okay, will wait for more info. Let me know if you think that we need this helper in SDK.
</blockquote></details>

Should not be a showstopper for this PR in any case (it's a "nice to have" and I could implement it later if needed).

---


*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/251)*

dennwc:
Okay, it seems like we are waiting only for @mcuadros review

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/bblfsh/sdk/251#-LAqlYtm2BsYBsK_TDyA:-LAsIbmw5PwDgcpFFGIF:b50u5hk) (no related file):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

Should not be a showstopper for this PR in any case (it's a "nice to have" and I could implement it later if needed).
</blockquote></details>

OK, marked as resolved for now.

---

*[a discussion](https://reviewable.io/reviews/bblfsh/sdk/251#-LAsIr727C5WtGhRJzum:-LAsIr727C5WtGhRJzun:b1b2o3x) (no related file):*
Before even thinking about merging this I would like to "pin" all current drivers to the current version of SDK with Dep, then merge this PR (old drivers will not be affected), and start merging new drivers.

@mcuadros @juanjux Do you have any other solutions in mind?

---


*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/251#-:-LAsNUp61rdxdYpX7Z0U:b7qfd5k)*

juanjux:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/bblfsh/sdk/251#-LAsIr727C5WtGhRJzum:-LAsTz_52tyF1D2PaiLN:b-4xjo4z) (no related file):*

Before even thinking about merging this I would like to "pin" all current drivers to the current version of SDK with Dep, then merge this PR (old drivers will not be affected), and start merging new drivers.

@mcuadros @juanjux Do you have any other solutions in mind?
</blockquote></details>

Yes, migrating the (current) drivers to Dep would be nice, but not really necessary, we could keep the "old" drivers using gopkg, publish the new SDK on gopkg as v2 (which it should be since it breaks backward compatibility) and the use dep on the new drivers.

---


*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/251)*

dennwc:
Review status: 63 of 64 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/bblfsh/sdk/251#-LAsIr727C5WtGhRJzum:-LAw5hVS08yz2IY6EAeG:b-lhahwu) (no related file):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

Yes, migrating the (current) drivers to Dep would be nice, but not really necessary, we could keep the "old" drivers using gopkg, publish the new SDK on gopkg as v2 (which it should be since it breaks backward compatibility) and the use dep on the new drivers.
</blockquote></details>

This was my initial proposal (to release `v2`), but @mcuadros mentioned that we don't want to support `v1` as a library, so we should update it directly. The only API compatibility promise right now is the driver response (protobuf).

In any case, I think it will be nice to add dep files to each driver to address https://github.com/bblfsh/sdk/issues/235 later. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/251)*

juanjux:
Review status: 63 of 64 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/bblfsh/sdk/251#-LAsIr727C5WtGhRJzum:-LAwp47v-mEkWWpmrRAY:befj4vi) (no related file):*

This was my initial proposal (to release `v2`), but @mcuadros mentioned that we don't want to support `v1` as a library, so we should update it directly. The only API compatibility promise right now is the driver response (protobuf).

In any case, I think it will be nice to add dep files to each driver to address https://github.com/bblfsh/sdk/issues/235 later. 
</blockquote></details>

Then... "dep" it is.

---


*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/251)*

dennwc:
Reviewed 1 of 1 files at r9, 2 of 2 files at r10.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io/reviews/bblfsh/sdk/251#-LAsIr727C5WtGhRJzum:-LBQT6TQ2zwyhVBPyWJ9:b-35qnt0) (no related file):*
<details><summary><i>Previously, juanjux (Juanjo Alvarez Martinez) wrote…</i></summary><blockquote>

Then... "dep" it is.
</blockquote></details>

Talked with Maximo, `sdk.v2` is okay, will merge right after the review.

---


*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/251)*

juanjux:
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/bblfsh/sdk/251#-LAsIr727C5WtGhRJzum:-LBQTy5aAZGdSjMvuxUx:bqfjazr) (no related file):*

Talked with Maximo, `sdk.v2` is okay, will merge right after the review.
</blockquote></details>

Perfect! 

---


*Comments from [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/251)*

protocol/types.go
```diff
@@ -187,3 +190,174 @@ type VersionResponse struct {
...
+	case 0:
+		return nil, nil
+	case 1:
+		return nd[0], nil
+	default:
+		return &Node{Children: nd}, nil
+	}
+}
+
+func asNode(n uast.Node, field string) ([]*Node, error) {
```
mcuadros: can you break this function in several ones, something like `arrayAsNode`, etc and call it from this function?

dennwc:
Awesome!

I will close this PR and manually push it as `v2` branch. Any additional commits will be done as PRs to that branch.

The reason not to merge to master directly is that we will need `gopkg.in` to see all new commits as soon as they appear. Once we have all beta drivers done and updated with latest DSL helpers, we will merge it into the master and finally tag it as `v2.0.0`.

PR #256, opened 2018-05-12
uast: define a new serialization format

dennwc:
Define a new serialization format for internal UAST representation.

Currently, it will only be used for tree diff project.


juanjux:
Shouldn't this wait to the meeting about the tree diff project?

dennwc:
@juanjux It's not directly related. This is just a simple binary serialization format that matches our new internal UAST format.

dennwc:
@mcuadros Again, as I mentioned in comments above, this is not a new set of types. We use similar types as internal UAST node representation in SDK v2, but convert to legacy format to pass it to bblfshd. This is just a way to serialize them.

The reason to have this kind of serialization is to pass raw UAST from SDK v2 to other tools like the upcoming Tree Diff.

Converting from SDKv2 to legacy and back loses some information. This loss includes:
- Types of properties, since legacy only supports string property values.
- Array types, since legacy can only describe it as an untyped node.
- Native AST node order, since legacy always sorts them by offset.

Also, the single array of children (as in legacy) is usually harder to process than smaller lists of children per property (as in SDKv2).

mcuadros:
The tree diff, should work with the agnostic interface, or directly based on our current UAST, I don't find any sense to make something based on an intermediate representation, not merged or hard to approve. 

The tree diff, should have a pattern similar to this other tree diff that we use:
https://github.com/src-d/go-git/tree/master/utils/merkletrie

dennwc:
@mcuadros I'm not sure why you call it an intermediate representation. It's the main representation of an SDKv2. And I already provided reasons why it's not useful for SDKv2 to have a diff on legacy trees.

An implementation of tree diff is currently in Python, so we need a way to serialize types from SDK to process it. This is exactly the reason to have the protobuf definition.

And thank you for the link, this pattern indeed hides all node details in a nice way. We will use the pattern once we decide to port tree diff prototype from Python.

dennwc:
Closing this, because this serialization is inefficient in any case. Will define a different (flat) format.

PR #258, opened 2018-05-16
Node comparison and load/store for multiple variables

dennwc:
* Add a node comparison function.
* Allow to load and store multiple variables from/to transformer state in one operation.


uast/node.go
```diff
@@ -222,6 +239,33 @@ func (m Object) EndPosition() *Position {
 	return AsPosition(o)
 }
 
+func (m *Object) SetNode(n Node) error {
```
juanjux: Need (and the others below) the doc comment.

Issue #253, opened 2018-04-20
Expose the listing drivers endpoint for clients


dpordomingo:
As required by https://github.com/bblfsh/client-scala/issues/68 and https://github.com/bblfsh/dashboard/issues/89, it would be needed to let the clients to fetch the list of installed drivers.

That API is part of the [bblfshd/daemon/protocol](https://github.com/bblfsh/bblfshd/blob/master/daemon/protocol/service.go#L15), and it's currently only [exported throwgh a unix socket for admin purposes](https://github.com/bblfsh/bblfshd/blob/master/cli/bblfshd/main.go#L51) so it can not be reached from the current clients.

Is there any plan to move the diver listing methods to the user network?

Here is the list of clients needing this:
- [ ] https://github.com/bblfsh/client-scala/issues/68 
- [ ] https://github.com/bblfsh/client-go/issues/53
- [ ] https://github.com/bblfsh/client-python/issues/98

juanjux:
Yes, Soon™

(Soon = as soon as we finish with the SDK and drivers migration).
--

dpordomingo:
As suggested by @bzz I added to the issue desc, a list of clients needing this.
(there is no issue yet in [client-python](https://github.com/bblfsh/client-python/issues))

zurk:
Now we have it for python client also https://github.com/bblfsh/python-driver/issues/155

dennwc:
@juanjux I can implement it. Don't want to block Apps because of this trivial feature.

juanjux:
@dennwc Talk with David, he was interested in doing it too.

bzz:
@dennwc 
>  Don't want to block Apps because of this trivial feature.

You are not - we actually have already something working for now, only for our case.

At the same time - we do not want to steal you precious time from more complex work and would be happy to contribute this one, under your careful review and supervision of course!

If that would be fine with you guys, @juanjux @dennwc  - AFAIK indeed @dpordomingo will be happy to start looking into that asap.

zurk:
@dpordomingo can you add https://github.com/bblfsh/client-python/issues/98 also in your list in first message please?

dpordomingo:
Added python to the desc.
I'd like to take a look at it this weekend and I'll ask you next week to confirm I understood how to do it (I never worked with bblfsh, so lets see if I can help :wink:

juanjux:
Excellent @dorpordomingo, btw we're hiring! 😁

dpordomingo:
[...]

Issue #167, opened 2017-08-30
Use an agglutinative language for Roles


abeaumont:
This would make the Roles language more expressive and flexible, remove complexity, make it more extensive and better suited for language analysis.

There are various tasks that need to be done for this:

- [x] Create a BIP describing the proposal: https://github.com/bblfsh/documentation/pull/82
- [x] Add versioning support for SDK
- [x] Modify the SDK roles #171
- [x] Update python driver to support new annotations
- [x] Update java driver to support new annotations https://github.com/bblfsh/java-driver/pull/47

abeaumont:
I think this can be closed now.

PR #259, opened 2018-05-30
sdk. Add SupportedLanguages method

dpordomingo:
Required by https://github.com/bblfsh/sdk/issues/253
Blocks: https://github.com/bblfsh/bblfshd/pull/167, https://github.com/bblfsh/client-go/pull/57, https://github.com/bblfsh/dashboard/pull/133

Since suported languages listing is not available through standard protocol it was added a `SupportedLanguages` method to get them.

These other PRs are blocked (needing) this implementation:
- `bblfsh/bblfshd`[#167](https://github.com/bblfsh/bblfshd/pull/167) implements `SupportedLanguages` method,
- `bblfsh/client-go`[#57](https://github.com/bblfsh/client-go/pull/57) exposes the `SupportedLanguagesRequest` and  `SupportedLanguagesResponse`,
- `bblfsh/dashboard`[#133](https://github.com/bblfsh/dashboard/pull/133) uses the `SupportedLanguages` feature exposed by `bblfsh/client-go`


protocol/service.go
```diff
@@ -31,6 +31,16 @@ func Version(req *VersionRequest) *VersionResponse {
 	return DefaultService.Version(req)
 }
 
+// SupportedLanguages uses DefaultService to process the given SupportedLanguagesRequest to get the supported drivers.
+//proteus:generate
+func SupportedLanguages(req *SupportedLanguagesRequest) *SupportedLanguagesResponse {
```
dennwc: `ListDrivers` might be a better name because we could potentially have multiple drivers for a single language (different versions/runtimes/forks).

protocol/types.go
```diff
@@ -187,3 +189,34 @@ type VersionResponse struct {
...
+//proteus:generate
+type SupportedLanguagesResponse struct {
+	Response
+	// Drivers are the loaded drivers
+	Drivers []DriverDetails `json:"drivers"`
+}
+
+// DriverDetails is the installed driver exported data
+//proteus:generate
+type DriverDetails struct {
```
dennwc: `DriverManifest`

sdk/driver/driver.go
```diff
@@ -153,3 +153,12 @@ func (d *Driver) Version(req *protocol.VersionRequest) *protocol.VersionResponse
 
 	return r
 }
+
+// SupportedLanguages handles a SupportedLanguagesRequest including information from the manifest.
+func (d *Driver) SupportedLanguages(req *protocol.SupportedLanguagesRequest) *protocol.SupportedLanguagesResponse {
+	return &protocol.SupportedLanguagesResponse{
+		Drivers: []protocol.DriverDetails{
```
dennwc: Not sure if we should allow drivers to list multiple languages...

juanjux: This is something to think about. For example, are python2 and 3, supported by the Python driver, different languages? (considering that they're incompatible). Also, a typescript driver could provide typescript and javascript.

protocol/service.go
juanjux: Agree.

protocol/types.go
```diff
@@ -187,3 +189,34 @@ type VersionResponse struct {
...
+	Name     string `json:"name"`
+	Language string `json:"language"`
+	Version  string `json:"version"`
+	Status   string `json:"status"`
```
juanjux: Could also have the `features` array of the manifest that gives a list of supported features.

protocol/service.go
dpordomingo: `ListDrivers` was my first option, but @mcuadros and @smola suggested `SupportedLanguages` because reasons.
I can rename it again if you agree :)

protocol/types.go
dpordomingo: I'll do so

dpordomingo: I just realized that to be consistent with `SupportedLanguages` name, this should renamed as `LanguagesManifest`, shouldn't it?
I'll rename it once we have a consensus about https://github.com/bblfsh/sdk/pull/259#discussion_r192006369 with @smola and @mcuadros.

protocol/types.go
```diff
@@ -187,3 +189,34 @@ type VersionResponse struct {
...
+// SupportedLanguagesRequest is a request to get the supported languages
+//proteus:generate
+type SupportedLanguagesRequest struct{}
+
+// SupportedLanguagesResponse is the reply to SupportedLanguagesRequest
+//proteus:generate
+type SupportedLanguagesResponse struct {
+	Response
+	// Drivers are the loaded drivers
+	Drivers []DriverDetails `json:"drivers"`
```
dpordomingo: I just realized that to be consistent with `SupportedLanguages` name, this should renamed as `Languages`, shouldn't it?
I'll rename it once we have a consensus about https://github.com/bblfsh/sdk/pull/259#discussion_r192006369 with @smola and @mcuadros.

sdk/driver/driver.go
dpordomingo: @mcuadros noticed me that a driver could stupor different languages.

protocol/service.go
juanjux: Let's see what the bosses say.

protocol/types.go
juanjux: features it's not the languages supported but the level of implementation (ast, uast, roles, and in the close future high level objects).

protocol/service.go
dennwc: It actually depends on the protocol level. For internal protocol `SupportedLanguages` makes sense. But for bblfshd `ListDrivers` is the only possible name (because of reasons mentioned above).

protocol/types.go
dennwc: I just don't see a reason why we should abstract away the real manifest that we have for the driver (this is what suggested by `SupportedLanguages` and omitting `features` field). A user can drop any unnecessary fields if he likes, but we won't be able to get missing fields if we decide to not include them.

dpordomingo: So you both @juanjux and @dennwc see a benefit if we include `features` field in the exposed data regardless the name of the method used?
If you agree I'll add it for sure :dagger:

protocol/service.go
smola: We had a discussion about what belongs to the administration API and what belongs to the public user API. I think we all agreed that listing supported languages is useful for end-users. We already have 4 very different use cases where this is good: bblfsh-playground, engine, gitbase, ml. Other aspects of driver install might or might not be needed.

@mcuadros also commented that public API should be the same as API for individual drivers standalone, but I think it's better that he elaborates that himself.

dennwc: I think that this API should cover most of the fields found in the driver manifest (although we can start by having only a minimal set of fields, like in current PR).

And I probably know the reason why Maximo wants an API to be exactly the same - it will allow using bblfshctl to send requests to individual drivers without additional flags to switch the protocol and without running bblfshd.

juanjux: After reading @smola then I think it could stay as it is and if in the future a driver supports more than one language we could add it to the returned SupportedLanguages here pointing to the same DriverDetails (or duplicating it) of the language supporting both.

protocol/types.go
dpordomingo: @juanjux I added `features` to the exposed data

dpordomingo:
@juanjux in case we go with the purposed `SupportedLanguages` and for the sake of the consistency, should I address one (or both) of the followings proposals before you merge this PR?
- rename `DriverDetails` as `DriverManifest` (following @dennwc [suggestion](https://github.com/bblfsh/sdk/pull/259#discussion_r191847522  )) or as `LanguagesManifest` (following my own [suggestion](https://github.com/bblfsh/sdk/pull/259#discussion_r192007552))
- rename `SupportedLanguagesResponse.Drivers` as `SupportedLanguagesResponse.Languages` as I [suggested](https://github.com/bblfsh/sdk/pull/259#discussion_r192007703))

juanjux:
@dpordomingo yes, I would choose DriversManifest as @dennwc said for the first option (because even if a driver provides several languages the manifest itself is a property of the driver, not the language), and SupportedLanguagesResponse.Languages.

dpordomingo:
@juanjux your [suggestions](https://github.com/bblfsh/sdk/pull/259#issuecomment-393856354) were addressed by e5abd94
I also added some tests cfb00d9 to keep code coverage over the current ratio.
Ping me if you need any other action on my side :dancer:

juanjux:
Awesome David, great work. @dennwc PTAL.

PR #263, opened 2018-06-18
Added info about protocol.Status

juanjux:
cc @smola

### DIFF
--- a/protocol/types.go
+++ b/protocol/types.go
@@ -62,7 +62,8 @@ const (
 	Base64
 )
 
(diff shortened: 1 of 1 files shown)


protocol/types.go
```diff
@@ -62,7 +62,8 @@ const (
 	Base64
 )
 
-// Response basic response, never used directly.
+// Response basic response, never used directly. The Response.Status field should always be
```
dennwc: Response is a ...

PR #265, opened 2018-07-02
Semantic objects

dennwc:
Merge semantic objects into SDK v2.


juanjux:
Sorry, I could only have a cursory look at this since I've been pretty busy with the remaining 1% (the harder) of the drivers but some things I wanted to comment before I forget:

- I don't think the `.native` fixtures should be changed to the new format because the purpose of these files is to check exactly what the output of the native driver is. For the `.uast` files it makes sense to use the new format since there is no "natural" representation of it and thus we can choose our own, but the native drivers output in a specified format and in this case I think being able to check it trumps readability.

- One thing I really miss when developing annotations and semantic objects is to have a "verbose/debug mode" from the SDK that would list, for every node, what annotations it try and why the ones that match do it and the ones that doesn't don't. But this could be added later.

dennwc:
@juanjux I don't think that changing native format from JSON to YML makes any difference. In both cases the driver actually decodes native JSON, so an original response is lost anyway. But I agree with you that it shouldn't add `@role` field for native output.

And I totally agree about the debug mode. But I'm still trying to find ways how to do it without dumping each node for each transform. If you have any ideas, please let me know.

In my private prototype a year ago I used Copy-on-Write for each operation and had a graph of nodes. By doing this I had a list of each node with its ID, all failed/passed transform for it and the history of node states after each transform for each node (since it used Copy-on-Write and all old states were preserved until graph GC). We might want to do the same for v3, but adding all of this into v2 will be overwhelming for reviewers ;).

And I totally agree regarding tree diff in tests. I will try to change it so it doesn't print the whole tree as a diff at least. But currently I'm not even trying to use that diff - it's always easier to use an external diff tool. This is why the test emits `_got` files - if the test fails you will see such files in the `./fixtures` directory and you can diff `foo.uast` (expected) with `foo.uast_got` (returned by the driver) to find the problem.

juanjux:
@dennwc let's move the discussion about the native format and other stuff and a debug mode with trace-level debugging to issues since neither are critical for release and we can merge this. The diff format can also be probably show without the `\n` and without the need for the user to run commands, but same, can be an improvement on next versions.

Awesome work by the way.

If you agree, I'll merge so we can start merging drivers.

dennwc:
@juanjux Sounds good, let's merge it! And thanks for the review :)

PR #267, opened 2018-07-11
Allow to specify a Github API token for driver discovery

dennwc:
Allow specifying a Github API token for driver discovery. Already added the token to our CI.


juanjux:
CI seems to be failing, but it could be github being slow.

PR #273, opened 2018-07-20
Implement XPath queries in SDK

dennwc:

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -29,10 +29,10 @@
   revision = "cd527374f1e5bff4938207604a14f2e38a9cf512"
 
 [[projects]]
-  name = "github.com/cenkalti/backoff"
+  branch = "master"
+  name = "github.com/antchfx/xpath"
   packages = ["."]
-  revision = "2ea60e5f094469f9e65adb9cd103795b73ae743e"
-  version = "v2.0.0"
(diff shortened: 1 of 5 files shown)


juanjux:
??

juanjux:
This can't be merged to the `~master` of `~v2` branches, reasons: not planned, not in the OKRs, not sufficiently discussed that we would move the XPath queries here, not tested that it'll work without problems with the Scala and Python clients (including memleaks, etc) and not tested with a good collection of the XPath queries that our users do. 

But it can be a separate branch for the moment until all that is resolved.

juanjux:
I would prefer all XPath queries to have the same backend to avoid implementation differences until we move to this. Could cause confusion for contributors to have two separate APIs for the same thing and for users to have potentially different XPath implementations.

We could collect a huge collection of queries that users have been using and if the results are the same and it's not slower we could add it as a private API inside the SDK, as an optimization. This could remove the "contributor confusion" part of the problem.

dennwc:
Then I propose to move it into `internal` package so no one except SDK itself can use it if you have concerns that someone may start relying on it. Once we test it well enough we can expose it to the public.

dennwc:
And I'm not proposing to use it for transformations. I was thinking of using it for fixture tests blacklists. Currently, we cannot blacklist based on some AST field, only by the node type. Might help to move blacklists to the preprocessor as well.

PR #276, opened 2018-07-20
Allow to run apt/apk in a final build

dennwc:

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -62,6 +62,7 @@ type buildManifest struct {
 	Native   struct {
 		Image  string     `yaml:"image"`
 		Static []artifact `yaml:"static"`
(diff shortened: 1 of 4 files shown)


build/build.go
```diff
@@ -96,8 +97,46 @@ func (d *Driver) readBuildManifest() (*buildManifest, error) {
...
 	}
+	for _, s := range m.Native.Deps {
+		if !isApkOrApt(s) {
+			return nil, fmt.Errorf("only apt/apk commands allowed in deps for final image")
+		}
+	}
 	return &m, nil
 }
+
+func isApkOrApt(s string) bool {
```
smola: As discussed previously, we'll restrict this at the moment, but we'll have further discussion about not restricting or restricting it further.

PR #277, opened 2018-07-20
Allow docker push on vX branches, but don't push latest

dennwc:
Allow docker push on `vX` branches, but don't push `latest` in this case.


juanjux:
Fix conflict please.

PR #278, opened 2018-07-24
Optimize UAST transforms

dennwc:
Address performance issues in the last SDK version.

In short: large transforms are an order of magnitude faster now. 

```
benchmark                         old ns/op      new ns/op     delta
BenchmarkGoDriver/transform-4     3181055629     261985232     -91.76%

benchmark                         old allocs     new allocs     delta
BenchmarkGoDriver/transform-4     18913347       561193         -97.03%

benchmark                         old bytes      new bytes     delta
BenchmarkGoDriver/transform-4     2509354920     54671790      -97.82%
```

### DIFF
--- a/uast/transformer/ops.go
+++ b/uast/transformer/ops.go
@@ -279,6 +279,21 @@ func (f FieldDescs) Clone() FieldDescs {
 	}
 	return fields
 }
+func (f FieldDescs) CheckObj(n nodes.Object) bool {
+	for k, d := range f {
+		if d.Optional {
+			continue
+		}
+		v, ok := n[k]
+		if !ok {
+			return false
+		}
+		if d.Fixed != nil && !nodes.Equal(*d.Fixed, v) {
+			return false
+		}
+	}
+	return true
+}
 
 // ObjectOp is an operation that is executed on an object. See Object.
 type ObjectOp interface {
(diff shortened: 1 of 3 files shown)


uast/transformer/ops.go
```diff
@@ -279,6 +279,21 @@ func (f FieldDescs) Clone() FieldDescs {
 	}
 	return fields
 }
+func (f FieldDescs) CheckObj(n nodes.Object) bool {
```
juanjux: Small doc comment missing.

uast/transformer/ops.go
```diff
@@ -458,7 +476,17 @@ func JoinObj(ops ...ObjectOp) ObjectOp {
...
 		required = nil
 	}
-	return opObjJoin{ops: out, partial: partial, allFields: required}
+	for i := 0; i < len(out); i++ {
+		op := out[i]
+		if len(op.fields) != 0 {
+			continue
+		}
+		if o, ok := op.op.(Obj); ok && len(o) == 0 && len(out) > 1 {
+			out = append(out[:i], out[i+1:]...)
```
juanjux: This could be potentially slow depending on how the runtime preallocates the slice.

dennwc: This happens only once on the program startup when the transforms are being built. So not so relevant.

PR #281, opened 2018-08-02
ast2gv: allow to omit position info and nil fields

dennwc:
AST2GV is a tool that converts a YML AST file to Graphviz/SVG/PNG image.

This PR adds two new CLI options:
* omit positional nodes,
* omit empty fields.

Currently, an AST might look a bit overloaded with information:

image

With both options enabled the result is much more readable:

image


smola:
@dennwc Looks good, but could you add a brief rationale on the commit message and/or PR description?

PR #282, opened 2018-08-02
Cleanup an old build system and unused commands

dennwc:
Since the introduction of a new build system, the old Makefiles are still included in the SDK.

PR removes these files, as well as old commands in `bblfsh-sdk-tools` used for the build.

It also fixes the wrong import for `pio` library.


dennwc:
@juanjux It's done, we can merge it now.

PR #287, opened 2018-08-10
Various fixes to init command

dennwc:


etc/skeleton/build.yml.tpl
```diff
@@ -2,15 +2,19 @@ sdk: '2'
 go-runtime:
   version: '1.10'
 native:
-  image: '{{.Language}}:latest'
+  image: 'debian:latest'
```
juanjux: Building native binaries with Debian is usually problematic when the binary is then run using Alpine because of the different libc, maybe we should default to Alpine.

dennwc: This is correct, but the new build system assumes that these images will be changed to `ruby-alpine`, for example, so these values are just placeholders to get a running dummy driver right after `bblfsh-sdk init`.

dennwc:
@smola Done!

etc/skeleton/build.yml.tpl
juanjux: So, could we use an alpine placeholder so an external driver programmer is not misguided into believing that debian based images are preferred?

dennwc: I can write a comment that Alpine is preferred, but forcing driver developers to deal with Alpine issues from the very beginning is a bit too hardcore.

Alpine doesn't have Bash so debugging driver is also not that nice. And cryptic errors for `libc` might puzzle anyone who is not familiar with Alpine.

Let's leave Debian as a default for new drivers since it's much easier to prototype things in.

juanjux: Yes, I understand, but cryptic errors regarding libc vs alpine libc on the final image is what I fear could happen there is they need to compile anything, maybe adding a comment like:

```bash
# If you need to natively compile, use an Alpine image instead
```

dennwc: The point is, they don't need Alpine up until they want to ship an optimized container which might happen after a few months of development. I don't see why we need to force the use of Alpine from day 1. It's totally okay to prototype the whole driver on Debian and if the driver starts to work properly at some point, then it's time to learn how to build on Alpine as the last step. But for the first step it's definitely an overkill.

Maybe our discussion started because you assume that the final docker image is always Alpine? Just to make sure I want to mention that all the steps in this `build.yml` are based on Debian:
`1.10` tag of Go is based on Debian, and all native stages (native build, test, final) is based on `debian:latest` according to the build file. No Alpine involved (yet).

dennwc: Sorry, but we are talking about two different things.

Yes, for all drivers built by source{d} and accepted as official will definitely use Alpine. Sure we were discussing that we might switch to `debian-slim` to simplify things, but the build system solved all the issues with Alpine. We will continue to use it internally, and it is the default for us.

But I want a short and simple blog post explaining how to build a driver from scratch, so we will get at least some initial contributions. I don't see any real reason why external people should jump directly into Alpine issues. And this is exactly what will happen if we set it as a default for `bblfsh-sdk init`. SDK is public, so let's make it a bit more user-friendly by default.

So I still insist on leaving the default value set to `debian` for `bblfsh-sdk init` and adding the comment to `build.yml` that the driver should be built on Alpine eventually.

juanjux: Adding a comment is what I indicated above, so yes, please add a coment noting that the final version should use Alpine and I'm ok with this.

PR #289, opened 2018-08-14
Fix Typo 

tsolakoua:
It's not part of the documentation, but found it randomly. 
Signed-off-by: Anna Tsolakou <tsolakou.a@gmail.com>


juanjux:
Thank you!

PR #294, opened 2018-08-21
Accept external node implementations

dennwc:
Since the new libuast version will be written in Go, there is a need to abstract UAST node type that is used in the code to allow to pass different implementations (implemented by client languages).

This CL defines a new `nodes.External` interface that will be implemented by native nodes, and checks for this kind of nodes in `ToNode`, so existing code will be able to handle external nodes at least by copying them.

It also updates `KindOf`, `Count` and `WalkPreOrder` functions to support external nodes instead of accepting only native ones.

### DIFF
--- a/uast/nodes/node.go
+++ b/uast/nodes/node.go
@@ -9,13 +9,19 @@ import (
 const applySort = false
 
 // Equal compares two subtrees.
-func Equal(n1, n2 Node) bool {
+// Equality is checked by value (deep), not by reference.
+func Equal(n1, n2 External) bool {
 	if n1 == nil && n2 == nil {
 		return true
-	} else if n1 != nil && n2 != nil {
-		return n1.Equal(n2)
+	} else if n1 == nil || n2 == nil {
+		return false
 	}
-	return false
+	if n, ok := n1.(Node); ok {
+		return n.Equal(n2)
+	} else if n, ok = n2.(Node); ok {
+		return n.Equal(n1)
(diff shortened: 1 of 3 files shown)


uast/nodes/node.go
```diff
@@ -9,31 +9,41 @@ import (
...
+		return n.Equal(n2)
+	} else if n, ok = n2.(Node); ok {
+		return n.Equal(n1)
 	}
-	return false
+	return equalExt(n1, n2)
 }
 
-// Node is a generic interface for a tree structure.
+// Node is a generic node interface.
```
juanjux: I think the previous definition was better since it doesn't use the same term it is defining.

juanjux:
How would this work? I see the External interface and the duplicated functions to handle it, but I don't understand how for example the Python client would implement it, since it's in Go.

dennwc:
This is only the Go interface definition that will be implemented by the libuast. You can check [this file](https://github.com/dennwc/sdk/blob/libuast/uast/libuast/nodes_c.go#L143) to see an actual implementation.

In short, Go will wrap a `NodeInterface` and the node handle provided by Python into a struct that implements the `External` interface. And the rest of SDK will work with it as if it was an ordinary Go implementation.

juanjux:
Thanks, the link and explanation adds the piece I was missing.

juanjux:
This would need some tests for the new methods and types.

PR #295, opened 2018-08-23
Allow to read raw UAST graph

dennwc:
Currently, SDK hides the fact that binary format for UAST assigns IDs to nodes. But, as was mentioned in the design doc and a recent [issue](https://github.com/bblfsh/web/pull/152), `uast-viewer` needs node IDs in the response.

This CL adds a new method to `nodeproto` package that allows to read a flat list of nodes and skip the tree reconstruction. This method will be used in the Dashboard server to send UASTs to the frontend.

### DIFF
--- a/uast/nodes/nodesproto/nodesproto.go
+++ b/uast/nodes/nodesproto/nodesproto.go
@@ -288,12 +288,59 @@ func (g *treeWriter) addNode(n nodes.Node) uint64 {
 	return id
 }
 
+// ReadTree reads a binary graph from r and tries to decode it as a tree.
+// If the graph is cyclic, an error is returned.
 func ReadTree(r io.Reader) (nodes.Node, error) {
 	g := newGraphReader()
 	if err := g.readGraph(r); err != nil {
 		return nil, err
 	}
 	return g.asTree()
+}
+
+type RawNode struct {
+	ID     uint64      `json:"id"`
+	Kind   nodes.Kind  `json:"kind"`
+	Value  nodes.Value `json:"val,omitempty"`
+	Keys   []uint64    `json:"keys,omitempty"`
+	Values []uint64    `json:"values,omitempty"`
+}
+
+type RawGraph struct {
+	Root  uint64             `json:"root,omitempty"`
+	Meta  uint64             `json:"meta,omitempty"`
+	Last  uint64             `json:"last,omitempty"`
+	Nodes map[uint64]RawNode `json:"nodes"`
+}
+
+// ReadRaw reads a graph from a binary stream, returning a flat list of all nodes.
+func ReadRaw(r io.Reader) (*RawGraph, error) {
+	g := newGraphReader()
+	if err := g.readGraph(r); err != nil {
+		return nil, err
+	}
+	rg := &RawGraph{
+		Root: g.root, Meta: g.meta, Last: g.last,
+		Nodes: make(map[uint64]RawNode, len(g.nodes)),
+	}
+	for id, n := range g.nodes {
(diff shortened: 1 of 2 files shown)


juanjux:
cc @smacker PTAL and check if this would serve you for the intended use case.

uast/nodes/nodesproto/nodesproto.go
```diff
@@ -433,6 +483,33 @@ func (g *graphReader) asTree() (nodes.Node, error) {
...
+		v, _ := asValue(m)
+		return nodes.KindOf(v)
+	}
+	if len(m.Keys) != 0 || m.IsObject {
+		return nodes.KindObject
+	}
+	return nodes.KindArray
+}
+
+func asValue(n *Node) (nodes.Value, error) {
```
smacker: Is it possible to make this method public?

uast/nodes/nodesproto/nodesproto.go
```diff
@@ -296,6 +298,54 @@ func ReadTree(r io.Reader) (nodes.Node, error) {
...
+	Value  nodes.Value `json:"val,omitempty"`
+	Keys   []uint64    `json:"keys,omitempty"`
+	Values []uint64    `json:"values,omitempty"`
+}
+
+type RawGraph struct {
+	Root  uint64    `json:"root,omitempty"`
+	Meta  uint64    `json:"meta,omitempty"`
+	Last  uint64    `json:"last,omitempty"`
+	Nodes []RawNode `json:"nodes"`
```
smacker: any reason to have it as an array instead of the map?

RawNode keeps ids in `Keys` and `Values` but those are original ids.
Most probably to read them I can do
```go
for _, id := range n.Values {
  fmt.Println(gr.Nodes[id - gr.Root])
}
```
but it feels unnatural and error prune. 

or have I misunderstood something?

uast/nodes/nodesproto/nodesproto.go
dennwc: Why do you need it? The `RawNode` has a `Value` method that stores an output of this function.

uast/nodes/nodesproto/nodesproto.go
dennwc: You don't need to subtract root, just make a lookup by id.

Yeah, I was thinking about a map, but it will be unordered in this case.

uast/nodes/nodesproto/nodesproto.go
smacker: true. My bad.

uast/nodes/nodesproto/nodesproto.go
smacker: Hm. Sorry, I must missing something.

But the default id of root is 1 right? And this is an array. So `gr.Nodes[gr.Root]` would return me the next node after root, right? And the same for all other ids. The would have `id - 1` index in the array. That's why I subtract root.

Even in tests I see a node has `id=1` but stays on `0` element of the array,

dennwc: First, nodes are sparse, meaning that you can have nodes `1,7,13`, and array indexes will be `0,1,2`, so you can't use an ID as an array index and you cannot calculate indexes. The only guarantee is that nodes are ordered by ID. 

Second, RootID can be any valid node ID from that array. For example it can be `7`.

You see root as a node with `id=1` is only because encoder emits nodes in BFS order. But this order can be arbitrary.

dennwc: I will change it to a map, then. Requiring sorting is definitely less common.

PR #301, opened 2018-08-30
Move iterators to the nodes package and implement positional iterator

dennwc:
* move base iterators to nodes package

* implement ToNode, TypeOf and NodeAs in terms of external nodes

* implement positional iterator

### DIFF
--- a/uast/iter.go
+++ b/uast/iter.go
@@ -1,0 +1,97 @@
+package uast
+
+import (
+	"sort"
+
+	"gopkg.in/bblfsh/sdk.v2/uast/nodes"
+)
+
+// NewPositionalIterator creates a new iterator that enumerates all object nodes, sorting them by positions in the source file.
+// Nodes with no positions will be enumerated last.
+func NewPositionalIterator(root nodes.External) nodes.Iterator {
+	return &positionIter{root: root}
+}
+
+type positionIter struct {
+	root  nodes.External
+	nodes []nodes.External
+}
+
+type nodesByPos struct {
+	nodes []nodes.External
+	pos   []Position
+}
+
+func (arr nodesByPos) Len() int {
+	return len(arr.nodes)
+}
+
+func (arr nodesByPos) Less(i, j int) bool {
+	return arr.pos[i].Less(arr.pos[j])
+}
+
+func (arr nodesByPos) Swap(i, j int) {
+	arr.nodes[i], arr.nodes[j] = arr.nodes[j], arr.nodes[i]
+	arr.pos[i], arr.pos[j] = arr.pos[j], arr.pos[i]
+}
+
+func (it *positionIter) sort() {
+	// in general, we cannot expect that parent node's positional info will include all the children
+	// because of this we are forced to enumerate all nodes and sort them by positions on the first call to Next
+	var plist []Position
+	noPos := func() {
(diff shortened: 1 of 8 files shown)


uast/iter.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (it *posIterator) sort() {
+	// in general, we cannot expect that parent node's positional info will include all the children
+	// because of this we are forced to enumerate all nodes and sort them by positions on the first call to Next
+	var plist []Position
+	noPos := func() {
+		plist = append(plist, Position{})
+	}
+	posType := TypeOf(Positions{})
+	nodes.WalkPreOrderExt(it.root, func(n nodes.External) bool {
```
juanjux: Can't this use the existing PreOrder iterator?

uast/iter.go
```diff
@@ -0,0 +1,97 @@
...
+	"gopkg.in/bblfsh/sdk.v2/uast/nodes"
+)
+
+// NewPositionalIterator creates a new iterator that enumerates all object nodes, sorting them by positions in the source file.
+// Nodes with no positions will be enumerated last.
+func NewPositionalIterator(root nodes.External) nodes.Iterator {
+	return &posIterator{root: root}
+}
+
+type posIterator struct {
```
juanjux: Should be positionIterator to be in line with the other rename and avoid being easy to mistype as postIterator (or vice-versa).

uast/iter_test.go
```diff
@@ -0,0 +1,54 @@
...
+			Name:    "B",
+		}),
+		toNode(Identifier{
+			GenNode: pos(0, 0, 0),
+			Name:    "C",
+		}),
+	}
+	a, b, c := root[0], root[1], root[2]
+	it := NewPositionalIterator(root)
+	expect(t, it, b, a, c)
```
juanjux: Shouldn't this be `c, b, a`?

uast/nodes/iter.go
```diff
@@ -0,0 +1,213 @@
...
+func (Empty) Next() bool     { return false }
+func (Empty) Node() External { return nil }
+
+type IterOrder int
+
+const (
+	IterAny = IterOrder(iota)
+	PreOrder
+	PostOrder
+	LevelOrder
```
juanjux: `PositionOrder` missing here.

uast/iter.go
dennwc: It will be impossible to skip nodes. This iterator also skips positional nodes.

uast/iter.go
dennwc: Mmm... It's an unexported type, so it doesn't matter that much. Personally (and following Go convention), I don't want long names in the method's receiver.

uast/iter_test.go
dennwc: "C" has an invalid position (there cannot be line 0). And all the nodes without positional info are sorted last.

uast/nodes/iter.go
dennwc: It's intentional - this package cannot define this iterator since all `Position`-related definitions are in the `uast` package. This enum is reflected into `query` package that has all kinds of iterators available.

dennwc:
Addressed all the feedback so I'm merging it

PR #303, opened 2018-09-12
Correct SDKv2 XPath projection

dennwc:
PR fixes the temporary XPath projection to follow SDKv2 spec.

The following node structure:
```json
{
  "@type": "uast:Alias",
  "Name": {
    "@type": "uast:Identifier",
    "Name": "A",
  },
  "Node": {
    "@type": "go:BasicLit",
    "@role": [ "Literal", "String" ],
    "@token": "abc",
    "Type": "STRING"
  }
}
```

Corresponds to the following XML structure in the new projection:
```xml
<uast:Alias>
  <Name>
    <uast:Identifier Name='A'></uast:Identifier>
  </Name>
  <Node>
    <go:Ident role='Literal' role='String' Type='STRING'>abc</go:Ident>
  </Node>
</uast:Alias>
```

Thus, a XPath for getting an identifier in Semantic UAST will be `//uast:Identifier` and the same query for Annotated UAST will be `//*[@role='Identifier']`.

There is also a change related to field names. In this new projection, field names are mapped to separate XML elements that wrap the value of the field. An XPath to get a specific field might look like this: `//uast:Alias/Name/*`.


uast/query/xpath/query.go
```diff
@@ -2,66 +2,104 @@ package xpath
...
+	// objectNode is an element.
+	objectNode
+	fieldNode
+	// valueNode is the text content of a node.
+	valueNode
+)
+
+type attr struct {
+	k string
+	v string
```
juanjux: I guess` key` and `value` are too long for the Go (sub)standards :)

uast/query/xpath/query.go
```diff
@@ -2,66 +2,104 @@ package xpath
...
+
+	n    nodes.External
+	kind nodes.Kind
+	obj  nodes.ExternalObject
+
+	tag   [2]string
+	attrs []attr
+	sub   []*node
+	par   *node
+	pari  int // index in parent's sub array
```
juanjux: paridx?

uast/query/xpath/query.go
```diff
@@ -2,66 +2,104 @@ package xpath
...
 }
 
 // nodeNavigator is for navigating JSON document.
 type nodeNavigator struct {
 	root, cur *node
+	attri     int
```
juanjux: What is `attri`?

uast/query/xpath/query.go
```diff
@@ -2,66 +2,104 @@ package xpath
...
+
+type node struct {
+	typ nodeType
+
+	n    nodes.External
+	kind nodes.Kind
+	obj  nodes.ExternalObject
+
+	tag   [2]string
+	attrs []attr
```
juanjux: Shouldn't this be a map?

dennwc:
Also, some results from a comparative benchmarks in go-client, running the same query over a large UAST:
```
BenchmarkXPathV1-4     10    166194261 ns/op     2610198 B/op      18037 allocs/op
BenchmarkXPathV2-4     30     53743641 ns/op    35384852 B/op     394715 allocs/op
```
It's ~3x times faster with no optimizations!

juanjux:
Those benchmark results are really nice. LGTM!

PR #305, opened 2018-09-18
Implement node hashing

dennwc:
Implement node hashing using SHA-256 and a simple serialization routine.

Note that it will hash the whole tree including positions. To omit positions, the caller will need to remove them from the tree.

### DIFF
--- a/uast/nodes/hash.go
+++ b/uast/nodes/hash.go
@@ -15,16 +15,36 @@ const HashSize = sha256.Size
 
 type Hash [HashSize]byte
 
+var DefaultHasher = NewHasher()
+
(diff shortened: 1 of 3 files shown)


juanjux:
I would prefer to have another Hashing method that doesn't take in account positions instead of forcing the user to remove them (which for most use cases it will). Or, alternatively, ignore positions by default.

dennwc:
This package cannot ignore positions because it has no knowledge of the schema.
Will add a `Hasher` type that will have a `KeyFilter func(string) bool` to skip positions.

dennwc:
@juanjux Done!

PR #306, opened 2018-09-19
Implement XPath filtering

dennwc:
The XPath implementation looks good, so it might be the right time to merge it to master. We can then start merging new libuast, client-go, etc that rely on XPath.


dennwc:
Okay, I'm assuming that everything is fine on the review side (was done before), so I'm merging it.

PR #307, opened 2018-09-19
Ignore Github rate limit errors in tests

dennwc:
Until we host a static manifest with the list of drivers, I propose to skip test failures if an error is due to rate limiting.

In any case, rate limiting affects 30-50% of our builds, thus the other 50-70% will be able to verify that the test passes (or not).

### DIFF
--- a/driver/manifest/discovery/discovery.go
+++ b/driver/manifest/discovery/discovery.go
@@ -196,6 +196,12 @@ type Options struct {
 	NoMaintainers bool   // do not load maintainers list
 }
 
+// isRateLimit checks if error is due to rate limiting.
+func isRateLimit(err error) bool {
+	_, ok := err.(*github.RateLimitError)
+	return ok
+}
+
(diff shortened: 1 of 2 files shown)


juanjux:
Mmmmm, but this could mask real errors if the test would fail but a rate limit error happens and then get merged. Could instead do some kind of wait and repeat (with a maximum number of retries) or it would be much of a hassle?

dennwc:
The problem is that SDK is not the reason for the rate limiting. It happens on the IP of an instance running in Travis. And of course, the instance get's heavily reused by others and some Travis clients also use Github API from time to time. I tried to place wait and retry logic, but it got an estimation from Github that the limit will be renewed after 1-3h on average.

And this piece of code is not changed too often. We will definitely spot if something goes wrong (we also have cron jobs on the repo).

juanjux:
I understand the problem, but still feel uneasy about this workaround skipping potentially failling tests, but I guess it'll be ok if we keep bblfsh/sdk/issues/264 open to eventually find a non-workaround solution.

PR #309, opened 2018-09-21
added UAST diff algorithm

quinor:
still WIP, diffing works, applying diff not implemented yet


uast/diff/diff.go
```diff
@@ -0,0 +1,283 @@
...
+		} else {
+			ds.push(&Attach{parent: parentID, key: parentKey, child: nil})
+		}
+
+	case matchDecision:
+		src, dst := src.(nodes.Object), dst.(nodes.Object)
+		keys := make(map[string]bool)
+		iterate := func(keyset nodes.Object) {
+			for key := range keyset {
+				if in := keys[key]; !in {
```
dennwc:
```golang
if in := keys[key]; in {
  continue
}
// ... rest of the code
```

uast/diff/diff.go
```diff
@@ -0,0 +1,283 @@
...
+			bestDecision.minEq(&decision{cost, sameDecision{}})
+		} else {
+			bestDecision.minEq(&decision{cost, matchDecision{}})
+		}
+
+	case nodes.Array:
+		src, dst := src.(nodes.Array), dst.(nodes.Array)
+		cost = 0
+		if len(src) == len(dst) && func() int {
+			sum := 0
```
dennwc: Make it a proper method of `ds`, or move the code out from an anonymous function.

uast/diff/diff.go
```diff
@@ -0,0 +1,283 @@
...
+	decisions map[keyType]decision
+	sizes     map[ID]int
+	changes   Changelist
+}
+
+func emptyCacheStorage() cacheStorage {
+	return cacheStorage{
+		make(map[keyType]decision),
+		make(map[ID]int),
+		make([]Change, 0),
```
dennwc: Please don't initialize empty slices like this - nil slices are perfectly fine.

Or even better, omit the field from initialization:
```golang
cacheStorage{
  decisions: make(...),
  sizes: make(...),
}
```

uast/diff/diff.go
```diff
@@ -0,0 +1,283 @@
+package diff
+
+import (
+	// TODO: integrate with https://github.com/src-d/lapjv if perf is unacceptable
```
dennwc: This TODO won't happen until `src-d/lapjv` is rewritten in pure Go. We won't bind to C in SDK.

uast/diff/diff.go
```diff
@@ -0,0 +1,283 @@
...
+	// one can always just create the dst ignoring the src
+	cost := ds.nodeSize(dst) + 1
+	bestDecision := decision{cost, replaceDecision{}}
+
+	if nodes.KindOf(src) != nodes.KindOf(dst) {
+		ds.decisions[label] = bestDecision
+		return bestDecision
+	}
+
+	switch src.(type) {
```
dennwc: If written as `switch src := src.(type) {`, the `src` variable will have the right type inside each `case`, so you won't need to cast it again.

uast/diff/diff.go
```diff
@@ -0,0 +1,283 @@
...
+
+	case nodes.Value:
+		src, dst := src.(nodes.Value), dst.(nodes.Value)
```
dennwc: Note that only `src` type assertion is safe. The second one will panic if `dst` is not a `Value`. Use `dst, ok := dst.(Value)` instead.

uast/diff/diff.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+
+		if len(src) != len(dst) {
+			cost = 2
+		}
+
+		if len(src) < len(dst) {
+			l := len(dst) - len(src)
+			for i := 0; i < l; i++ {
+				src = append(src, nil)
```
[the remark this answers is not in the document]
quinor: Fully aware. I'm not using ID(src) anywhere later.

uast/diff/diff.go
quinor: look 5 lines above:

```
	if nodes.KindOf(src) != nodes.KindOf(dst) {
		ds.decisions[label] = bestDecision
		return bestDecision
	}
```

uast/diff/diff.go
quinor: I wanted to do a switch with both src and dst but the go wouldn't let me. Both variables have the same type, I've checked that.

uast/diff/diff.go
quinor: I wasn't sure whether "won't" or "don't really want to" so left the reminder our impl exists

uast/diff/diff.go
quinor: it's used only once here and won't be used anywherte else. I've used the lambda only in order to leverage the lazy ifs (&& won't execute 2nd argument if 1st is false, right?) and avoid nested if construction. How to proceed then?

uast/diff/diff.go
```diff
@@ -0,0 +1,283 @@
...
+
+type decisionType interface {
+	isDecisionType()
+}
+
+type basicDecisionType struct{}
+
+func (_ basicDecisionType) isDecisionType() {}
+
+type decision struct {
```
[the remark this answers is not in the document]
quinor: done.

uast/diff/changelist.go
```diff
@@ -0,0 +1,58 @@
...
+
+type Change interface {
+	isChange()
+}
+
+type changeBase struct {
+	txID uint64
+}
+
+func (_ *changeBase) isChange() {}
```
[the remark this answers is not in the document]
quinor: changed to lack of pointer here and in a few more places

uast/diff/diff.go
```diff
@@ -0,0 +1,287 @@
...
+	"github.com/heetch/lapjv"
+	"gopkg.in/bblfsh/sdk.v2/uast/nodes"
+)
+
+type decisionType interface {
+	isDecisionType()
+	cost() int
+}
+
+type basicDecision struct {
```
dennwc: This can be also inlined to every decision type. Yes, it will require to write `cost()` function for each type, but it will be easier to work with decisions: `replaceDecision{cost}` instead of `replaceDecision{basicDecision{cost}}`.

uast/diff/diff.go
```diff
@@ -0,0 +1,287 @@
+package diff
+
+import (
+	"fmt"
+	"github.com/heetch/lapjv"
+	"gopkg.in/bblfsh/sdk.v2/uast/nodes"
+)
+
+type decisionType interface {
+	isDecisionType()
```
dennwc: `isDecisionType()` can be removed, since `cost()` already prevents external packages from implementing it (assuming other types in this package has no such method).

uast/diff/diff.go
```diff
@@ -0,0 +1,287 @@
...
+// match decision types together with their params
+type sameDecision struct{ basicDecision }
+type replaceDecision struct{ basicDecision }
+type matchDecision struct{ basicDecision }
+type permuteDecision struct {
+	basicDecision
+	permutation []int
+}
+
+//convinience method for choosing the cheapest decision
```
dennwc: In Go, all comments should start with a name of a function/type:
> min is a convenience ...

uast/diff/diff.go
```diff
@@ -0,0 +1,287 @@
...
+type keyType struct{ k1, k2 ID }
+
+//cache for diff computation
+type cacheStorage struct {
+	decisions map[keyType]decisionType
+	sizes     map[ID]int
+	changes   Changelist
+}
+
+func emptyCacheStorage() cacheStorage {
```
dennwc: `newCacheStorage` or `makeCacheStorage`
It should also return a pointer.

uast/diff/diff.go
```diff
@@ -0,0 +1,287 @@
...
+	return cacheStorage{
+		make(map[keyType]decisionType),
+		make(map[ID]int),
+		nil,
```
dennwc: Again, please don't initialize things with `nil`. It's not C - Go takes care of memory initialization with zero values.
Use named fields instead:
```
&cacheStorage{
  decisions: make(map[keyType]decisionType),
  sizes: make(map[ID]int),
}
```

uast/diff/diff.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+}
+
+//type for cache
+type keyType struct{ k1, k2 ID }
+
+//cache for diff computation
+type cacheStorage struct {
+	decisions map[keyType]decisionType
+	sizes     map[ID]int
```
dennwc: Maybe rename to `counts`? `sizes` forces me to think it's the size of an array or an object (non-recursive).

uast/diff/diff.go
```diff
@@ -0,0 +1,287 @@
...
+
+// find the cheapest way to naively match src and dst and return the action with its combined cost
+func (ds *cacheStorage) decideAction(src, dst nodes.Node) decisionType {
+	label := keyType{nodes.UniqueKey(src), nodes.UniqueKey(dst)}
+
+	if val, ok := ds.decisions[label]; ok {
+		return val
+	}
+
+	// one can always just create the dst ignoring the src
```
dennwc: Please elaborate what this has to do with the code below.

uast/diff/diff.go
dennwc: Comment is still valid - you can write `src := src.(type)` and drop type assertions on `src` (but not `dst`).

uast/diff/diff.go
```diff
@@ -0,0 +1,287 @@
...
+		iterate := func(keyset nodes.Object) {
+			for key := range keyset {
+				if in := keys[key]; !in {
+					keys[key] = true
+					cost += ds.decideAction(src[key], dst[key]).cost()
+				}
+			}
+		}
+		iterate(src)
+		iterate(dst)
```
dennwc: You are not resetting `keys` here. Is it intentional?

Also, you can call `Keys()` on both objects and scan these slices (they are sorted by key).

uast/diff/diff.go
dennwc: I would say it should be a separate method on `ds` then. Inlining functions like this is not an idiomatic Go.

uast/diff/diff.go
```diff
@@ -0,0 +1,287 @@
...
+		if len(src) != len(dst) {
+			cost = 2
+		}
+
+		if len(src) < len(dst) {
+			arr := make(nodes.Array, len(dst))
+			copy(arr, src)
+			src = arr
+		} else {
+			arr := make(nodes.Array, len(src))
```
dennwc: This branch will trigger for the case of `len(src) == len(dst)` and different children (`cost != 0`).

uast/diff/diff.go
```diff
@@ -0,0 +1,287 @@
...
+			copy(arr, dst)
+			dst = arr
+		}
+		n := len(src)
+		m := make([][]int, n)
+		for i := 0; i < n; i++ {
```
dennwc: `for i := range m` and the same for `j`

uast/diff/diff.go
```diff
@@ -0,0 +1,287 @@
...
+		return
+	}
+	ds.push(Create{Node: node})
+}
+
+func (ds *cacheStorage) generateDifference(src, dst nodes.Node, parentID ID, parentKey Key) {
+	decision := ds.decideAction(src, dst)
+	switch d := decision.(type) {
+
+	//no action required if same
```
dennwc: move it inside the `case` block

uast/diff/diff.go
dennwc: This comment still applies

uast/diff/diff.go
```diff
@@ -0,0 +1,287 @@
...
+					keys[key] = true
+					if _, ok := dst[key]; !ok {
+						ds.push(Deattach{Parent: nodes.UniqueKey(src), Key: String(key)})
+					} else {
+						ds.generateDifference(src[key], dst[key], nodes.UniqueKey(src), String(key))
+					}
+				}
+			}
+		}
+		iterate(src)
```
dennwc: Same as above - `keys` is not wiped and there is another way of doing it with less iterations.

uast/diff/diff.go
```diff
@@ -0,0 +1,287 @@
...
+		for i1, i2 := range d.permutation {
+			if i1 != i2 {
+				recreate = true
+				break
+			}
+		}
+		if recreate {
+			//recreate src with right perm
+			newsrc := make([]nodes.Node, len(dst))
+			for i := 0; i < len(dst); i++ {
```
dennwc: `for i := range dst`

uast/diff/diff.go
```diff
@@ -0,0 +1,287 @@
...
+				newsrc[i] = src[d.permutation[i]]
```
dennwc:
```
newsrc := make([]nodes.Node, 0, len(dst))
```
and
```
newsrc = append(newsrc, src[d.permutation[i]])
```

uast/diff/diff.go
```diff
@@ -0,0 +1,287 @@
...
+		}
+		for i1, i2 := range d.permutation {
+			if i1 != i2 {
+				recreate = true
+				break
+			}
+		}
+		if recreate {
+			//recreate src with right perm
+			newsrc := make([]nodes.Node, len(dst))
```
dennwc: It's confusing that `newsrc` has the length of `dst` and iterates over `dst` index, but accesses `src` and `d.permutation` instead. If I'm not mistaken, here `src` has the same size, so `len(src)` can be used instead.

dennwc:
Also, it would be nice to have some tests for diff generation as well as for appling the diff.

uast/diff/apply.go
```diff
@@ -0,0 +1,68 @@
...
+		switch ch := change.(type) {
+		case Create:
+			//create a node and add to the dictionary
+			nodeDict[nodes.UniqueKey(ch.Node)] = ch.Node
+
+		case Attach:
+			//get src and chld from the dictionary, attach (modify src)
+			parent, ok := nodeDict[ch.Parent]
+			if !ok {
+				panic("I don't know what to get attached to!")
```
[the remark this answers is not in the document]
quinor: Those were temporary and slipped when I committed somehow...

uast/diff/diff.go
quinor: done.

uast/diff/changelist.go
```diff
@@ -0,0 +1,60 @@
...
+	Parent ID
+	Key    Key
+	Child  ID
+}
+
+// deattach a child from a node
+type Deattach struct {
+	changeBase
+	Parent ID
+	Key    Key //or string, how to do alternative?
```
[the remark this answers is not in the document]
quinor: I am not sure whether deattach operation makes sense for an array. It could have semantics of making such child a null for example.

uast/diff/diff.go
quinor: not resetting keys is the point. I want to visit every key from the joint key set exactly once. Haven't found a better method in Go.

quinor: What is Keys() function? Can't find any docs for it.

Also, added better comments.

uast/diff/diff.go
quinor: right. fixed.

uast/diff/diff.go
quinor: (see the other discussion)

uast/diff/diff.go
quinor: No. `src` is/can be of different length than `dst`, that's why I'm creating a `newsrc` with length of `dst` and elements of `src` permuted by `d.permutation`.

quinor:
How do we add tests in the repository? I've got some internal ones.

dennwc:
@quinor You can add a test function to a new file `./uast/diff/diff_test.go`. CI will call it automatically.

quinor:
@dennwc how to proceed if the test(s) need serialized data or external resources to work? My current testing procedure is to launch a client that runs the diff on two files and tries to apply it afterwards comparing the result with source, additionally having bblfshd docker running for the source file parsing.

dennwc:
@quinor Feel free to add a `fixtures` or `testdata` folder and add few YAML UAST files to it. Tests can read and diff those files. No need to run the whole client+bblfshd chain.

To get those YAMLs you can run the `bblfsh-cli` included in the latest Go client. It can connect to bblfshd, parse files and emit the YAML/binary formats.

quinor:
Is there any documentation/examples for the go+CI we are using?

dennwc:
@quinor Don't think about CI right now, you can use `go test ./...` to test it locally. The same command will be executed by CI.

quinor:
Did not realize there is an unified "go test" framework. That simplifies things alot :)

uast/diff/changelist.go
```diff
@@ -0,0 +1,60 @@
+package diff
+
+import (
+	"gopkg.in/bblfsh/sdk.v2/uast/nodes"
+)
+
+type Changelist []Change
+
+type Change interface {
+	isChange()
```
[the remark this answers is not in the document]
dennwc: No, this method exists to emulate type unions (a common practice in Go).

uast/diff/diff.go
dennwc: `Keys` is a function on `nodes.Object`

uast/diff/diff.go
```diff
@@ -0,0 +1,296 @@
...
+			for i := range src {
+				sum += ds.decideAction(src[i], dst[i]).cost()
+			}
+			if sum == 0 {
+				bestDecision = min(bestDecision, sameDecision{basicDecision{cost}})
+				break
+			}
+		}
+
+		if len(src) != len(dst) {
```
dennwc:
```suggestion
		} else {
```

uast/diff/diff.go
dennwc: Answered above

uast/diff/testdata/create_testdata.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python3
+
+# This is a python script used to generate test data for the diff library. It uses pairs of files
+# to get diffed acquired from
+# https://github.com/vmarkovtsev/treediff/blob/49356e7f85c261ed88cf46326791765c58c22b5b/dataset/flask.tar.xz
+# It uses https://github.com/bblfsh/client-go#Installation to convert python sources into
+# uast binary files.
```
dennwc:
```suggestion
# uast yaml files.
```

uast/diff/testdata/create_testdata.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python3
```
dennwc: The script is pretty simple. Can be written in Go as well (it's the only Python file here).

uast/diff/testdata/smalltest.txt
```diff
@@ -0,0 +1,148 @@
+720_1
```
dennwc: Maybe there is a formal definition for this? For example "source size is less than N bytes/lines" or "the UAST size is less than N bytes/nodes"?

uast/diff/testdata/create_testdata.py
```diff
@@ -0,0 +1,36 @@
...
+    return glob("{}/{}_before*.src".format(pwd, name))[0]
+
+
+def get_dst(name):
+    return glob("{}/{}_after*.src".format(pwd, name))[0]
+
+i = 0
+for src, dst in ((get_src(name), get_dst(name)) for name in testnames):
+    print(i, src)
+    os.system("bblfsh-cli -l python {} -o yaml > testcase_{}_src.uast".format(src, i))
```
dennwc: Assuming the test is changed to accept `*_src.uast`:
```suggestion
    os.system("bblfsh-cli -l python {} -o yaml > {}_src.uast".format(src, i))
```

uast/diff/testdata/smalltest.txt
quinor: I'm not sure how I'd chosen these. If there was such definition, how to proceed?

uast/diff/apply.go
```diff
@@ -0,0 +1,68 @@
+package diff
+
+import (
+	"fmt"
+	"gopkg.in/bblfsh/sdk.v2/uast/nodes"
+)
+
+func Apply(root nodes.Node, changelist Changelist) nodes.Node {
```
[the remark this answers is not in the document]
quinor: done

uast/diff/testdata/create_testdata.py
quinor: I only threw it in so that one may try to regenerate/generate new test cases. Also, see the answers below.

uast/diff/testdata/smalltest.txt
quinor: Also, Python exists on every modern Linux distro so it's still quite portable (see the alternative: bash)

uast/diff/testdata/create_testdata.py
dennwc: I still think it makes sense to rewrite it in Go. All the SDK is in Go. I see no reason to require Python installation for such a simple script.

dennwc:
Needs a rebase on top of a current master at least.

Also make sure to run `go fmt ./...`

dennwc:
Have to close this PR - cannot rebase, because it's made from the `master` branch on the fork. See https://github.com/bblfsh/sdk/pull/396.

PR #316, opened 2018-10-12
bblfsh-sdk-tools: move all commands to bblfsh-sdk

bzz:
Based on #316, so only 04444e8 is new.

Following the discussion https://github.com/bblfsh/sdk/pull/313#pullrequestreview-164171051 this is a proposal to move `$ bblfsh-sdk-tools fistures` and `$ bblfsh-sdk-tools ast2gv` to the single CLI app `$ bblfsh-sdk` and removes `bblfsh-sdk-tool`.


bzz:
@juanjux as this was based on #316  - all changes got in now :)

I would expect to merge 316 first and then rebase this to only include last commit, sorry for confusion!

Issue #257, opened 2018-05-14
Clarify error statuses returned by native drivers


dennwc:
Babelfish drivers may return 3 main statuses: `OK`, `Error`, `Fatal`.

Currently, there is no guarantee that `Error` responses will have a partial AST, since it's not tested in any of the drivers.

At the same time, `Error` status is not considered an error/exception in clients as reported in https://github.com/bblfsh/java-driver/issues/77, thus clients might not be aware that they receive a partial AST.

We should either:
a) Make a special type of error/exception that user code should assert to accept a partial AST. In this case clients will receive an error in case of any syntax erros, which might be a desired behavior.
b) Mention current behavior in Babelfish docs, clarify that user should check the status.

Personally, I would prefer the first option.

dennwc:
Ping @smola @juanjux

juanjux:
It's documented that the user should check the status code in the [Reading and Interpreting the Response](https://doc.bblf.sh/user/server-grpc-example.html#reading-and-interpreting-the-response) section.

dennwc:
> ...only a value of `protocol.Status.OK` will indicate success.

This is what is written in docs. I think this means that clients should raise an error/exception for all statuses other than `OK` and users will not be forced to check it.

juanjux:
The full quote, emphasis mine: 

> **You should check the status** (Status in the case of Go, since public members start with uppercase); only a value of protocol.Status.OK will indicate sucess.

I think it's clear in saying that the status should be checked. It doesn't say that the clients should raise an exception. 

Once that's said, can talk about if we want the clients to raise an exception for the `Error` status. Personally, I think a broken source file is not an exceptional failure (especially when you are parsing thousands and exceptions in Scala can have a performance penalty unlike in Python) and clients should check the status (it's there for a reason), but if other people and most importantly our existing users agree we could change that.

juanjux:
Could be an Optional in Scala which AFAIK doesn't have the performance penalty of exceptions and force the user to check them before usage.

smola:
Scala's `Option` doesn't work here, since you could not check the error message in case of error. `Option` can be `Some(result)` or `None`.

You have other options to avoid exceptions:

Define a `DriverException` with a method to retrieve the failed `UAST`, and then one of the following:

[...]

PR #321, opened 2018-10-23
Define and propagate driver error codes

dennwc:
Define common errors for native drivers and propagate them properly via gRPC boundary.

The list of errors so far:

- Syntax error (`ErrSyntax`): either partial or no UAST is returned. In gRPC, the status is `OK`, but an `Errors` field of the response contains error descriptions. 

- Native driver failure (`ErrDriverFailure`): native driver crashed or the native protocol connection is broken.  This indicates a bug in the driver and should be reported. Unrecoverable when returned by the driver. Driver server (either Go or `bblfshd`) should attempt a native driver restart before returning this error to the client.

- Transformation failed (`ErrTransformFailure`): parsing was succesful, but transformation pipeline for selected `Mode` fails to process it. This indicates a bug in the driver and should be reported. Recoverable, assuming `Mode` parameter is changed.

- All other errors are considered protocol/server failures.

Related:
- https://github.com/bblfsh/sdk/issues/257
- https://github.com/bblfsh/client-go/issues/82

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -75,6 +75,9 @@ type ParseOptions struct {
 type Driver interface {
 	// Parse reads the input string and constructs an AST representation of it.
 	//
+	// Language can be specified by providing ParseOptions. If the language is not set,
+	// it will be set during the Parse call if implementation supports language detection.
+	//
 	// Depending on the mode, AST may be transformed to different UAST variants.
 	// ErrModeNotSupported is returned for unsupported transformation modes.
 	//
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -65,6 +65,12 @@ type Module interface {
 	Close() error
 }
 
+type ParseOptions struct {
+	Mode     Mode
+	Language string
+	Filename string
+}
+
 // Driver is an interface for a language driver that returns UAST.
 type Driver interface {
 	// Parse reads the input string and constructs an AST representation of it.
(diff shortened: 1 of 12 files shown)


driver/driver.go
```diff
@@ -40,9 +65,28 @@ type Module interface {
...
+	//
+	// Depending on the mode, AST may be transformed to different UAST variants.
+	// ErrModeNotSupported is returned for unsupported transformation modes.
+	//
+	// Syntax errors are indicated by returning ErrSyntax.
+	// In this case a non-empty UAST may be returned, if driver supports partial parsing.
+	//
+	// Native driver failures are indicated by ErrDriverFailure and UAST transformation are indicated by ErrTransformFailure.
+	// All other errors indicate a protocol or server failure.
+	Parse(ctx context.Context, src string, opts *ParseOptions) (nodes.Node, error)
```
bzz: Could someone please help me understand the impact of this change: what else needs to be changed and where?

dennwc: This only affects an SDK itself. Drivers will also need to be updated, but only to bump version. They just call to SDK that creates and hosts those `Driver`s, thus no other changes are necessary. So I think it's safe to assume it's a minor change.

dennwc: Regarding "taking advantage" part, the `client-go.v3` needs to be updated (already have the change, but it's blocked by this). Again, only internals of the client will change. User-facing API will not be affected.

PR #324, opened 2018-10-31
Enable namespaces

dennwc:
The last missing piece of planned features for UASTv2. As mentioned in the design doc, native AST types should have namespaces matching the driver's language when the transformation is in Semantic mode. This will allow to avoid name collisions when analyzing ASTs for different languages.


dennwc:
Merging, since a similar change was already approved.

PR #328, opened 2018-11-06
Flag added to test cmd to disable syntax check

juanjux:
Adds `bblfsh-sdk test --nosyntaxcheck` flag for disabling the mandatory syntax check on drivers with syntax-error-proof parsers like C++ and Bash.


juanjux:
Pseudo duplicate of bblfsh/sdk/pull/327

We should choose one :)

PR #329, opened 2018-11-21
Shared gRPC server config between bblfshd/driver

bzz:
Part of the https://github.com/bblfsh/client-go/issues/102

Includes:
 - move `sdk.BuildGRPCOptions` from bblfshd to SDK
 - add same CLI `grpc-max-message-size` arg to all Drivers

### DIFF
--- a/cmd/util_test.go
+++ b/cmd/util_test.go
@@ -1,0 +1,25 @@
+package cmd
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestGRPCOptions_InvalidInput(t *testing.T) {
+	opts, err := GRPCSizeOptions(-1)
+
+	require.Error(t, err, "a given value was not applied")
+	require.NotNil(t, opts)
+	require.Len(t, opts, 2)
+	// does not work as expected for Func values like grpc.ServerOption
+	//require.Contains(t, opts, grpc.MaxRecvMsgSize(DefaulGRPCMaxSendRecvMsgSizeMB))
+	//require.Contains(t, opts, grpc.MaxSendMsgSize(DefaulGRPCMaxSendRecvMsgSizeMB))
+}
+
+func TestGRPCOptions_ValidInput(t *testing.T) {
+	opts, err := GRPCSizeOptions(32)
+
+	require.Nil(t, err)
+	require.NotNil(t, opts)
+}
--- a/cmd/util_test.go
+++ b/cmd/util_test.go
@@ -7,14 +7,15 @@ import (
 )
 
 func TestGRPCOptions_InvalidInput(t *testing.T) {
+	// too small
 	opts, err := GRPCSizeOptions(-1)
+	require.Error(t, err, "a small value should not be applied")
+	require.Nil(t, opts)
 
-	require.Error(t, err, "a given value was not applied")
-	require.NotNil(t, opts)
-	require.Len(t, opts, 2)
-	// does not work as expected for Func values like grpc.ServerOption
-	//require.Contains(t, opts, grpc.MaxRecvMsgSize(DefaulGRPCMaxSendRecvMsgSizeMB))
-	//require.Contains(t, opts, grpc.MaxSendMsgSize(DefaulGRPCMaxSendRecvMsgSizeMB))
+	// too big
+	opts, err = GRPCSizeOptions(maxMsgSizeCapMB + 1)
+	require.Error(t, err, "a big value should not be applied")
+	require.Nil(t, opts)
 }
 
 func TestGRPCOptions_ValidInput(t *testing.T) {
@@ -22,4 +23,8 @@ func TestGRPCOptions_ValidInput(t *testing.T) {
 
 	require.Nil(t, err)
 	require.NotNil(t, opts)
+	require.Len(t, opts, 2)
+	// does not work as expected for Func values like grpc.ServerOption
+	//require.Contains(t, opts, grpc.MaxRecvMsgSize(DefaulGRPCMaxSendRecvMsgSizeMB))
+	//require.Contains(t, opts, grpc.MaxSendMsgSize(DefaulGRPCMaxSendRecvMsgSizeMB))
 }
--- a/common.go
+++ b/common.go
@@ -15,29 +15,28 @@ const (
 	// MaxMsgSizeCLIDesc is description for the CLI flag to set max msg size.
 	MaxMsgSizeCLIDesc = "max. message size to send/receive to/from clients (in MB)"
 
-	// DefaulGRPCMaxMsgSizeMb is maximum msg size for gRPC in Mb.
-	DefaulGRPCMaxMsgSizeMb = 100
-	gRPCMaxMsgSizeCapMb    = 2048
+	// DefaulGRPCMaxSendRecvMsgSizeMB is maximum msg size for gRPC in MB.
+	DefaulGRPCMaxSendRecvMsgSizeMB = 100
+	maxMsgSizeCapMB                = 2048
 )
 
-// GRPCOptions returns a slice of gRPC server options with the maximum
-// message size set for sending and receiving.
-// Is intended to be shared by gRPC in bblfshd Server and Drivers.
-// Sets the hard limit of message size to less than 2GB since
-// it may overflow an int value, and it should be big enough.
-func GRPCOptions(maxMessageSizeMb int) ([]grpc.ServerOption, error) {
+// GRPCOptions returns a slice of gRPC server options with the max
+// message size the server can send/receive set.
+// If a >2GB value is requested: the maximum size limit is capped
+// at 100 MB and an error is returned.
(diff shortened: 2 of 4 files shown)


common.go
```diff
@@ -1,4 +1,42 @@
...
 const NativeBin = "/opt/driver/bin/native"
 const NativeBinTest = "/opt/driver/src/build/native"
+
+const (
+	// MaxMsgSizeCLIName is name of the CLI flag to set max msg size.
+	MaxMsgSizeCLIName = "grpc-max-message-size"
+	// MaxMsgSizeCLIDesc is description for the CLI flag to set max msg size.
+	MaxMsgSizeCLIDesc = "max. message size to send/receive to/from clients (in MB)"
+
+	// DefaulGRPCMaxMsgSizeMb is maximum msg size for gRPC in Mb.
```
creachadair: nit: please be consistent about "Mb" vs "MB". "Mb" usually means "megabits", whereas here it appears to be a capitalization wart maybe(?)

Ideally: "MiB" if we really mean powers of 2.

creachadair: Relatedly: For the gRPC request/reply? Or the whole size including the envelope?

common.go
```diff
@@ -1,4 +1,42 @@
...
+)
+
 const NativeBin = "/opt/driver/bin/native"
 const NativeBinTest = "/opt/driver/src/build/native"
+
+const (
+	// MaxMsgSizeCLIName is name of the CLI flag to set max msg size.
+	MaxMsgSizeCLIName = "grpc-max-message-size"
+	// MaxMsgSizeCLIDesc is description for the CLI flag to set max msg size.
+	MaxMsgSizeCLIDesc = "max. message size to send/receive to/from clients (in MB)"
```
creachadair: Does this need to be exported?

creachadair: Oh, nevermind—I see this is for setting flag strings.

It is weird to put this into the SDK. This may be a bigger design question, so feel free to defer it, but if we are going to let libraries choose labels for command-line parameters, we shouldn't export the strings and force the caller to do the right thing—instead, we should give the library a function the caller can use to generate flags. 

But there is probably some established convention here, I'm still learning my way around.

common.go
```diff
@@ -1,4 +1,42 @@
...
+
+	// DefaulGRPCMaxMsgSizeMb is maximum msg size for gRPC in Mb.
+	DefaulGRPCMaxMsgSizeMb = 100
+	gRPCMaxMsgSizeCapMb    = 2048
+)
+
+//BuildGRPCOptions creats gRPC ServerOption \w maxRecv/Send msg size set.
```
creachadair: nit: space after `//`

Suggestion (https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences):

```
// GRPCOptions returns a slice of gRPC server options with the maximum
// message size set for sending and receiving.
```

common.go
```diff
@@ -1,4 +1,42 @@
...
+// Is intended to be shared by gRPC in bblfshd Server and Drivers.
+// Sets the hard limit of message size to less than 2GB since
+// it may overflow an int value, and it should be big enough.
+func BuildGRPCOptions(maxMessageSizeMb int) ([]grpc.ServerOption, error) {
```
creachadair: "Build" isn't adding anything here. `sdk.GRPCOptions` seems sufficient.

Or if there's more than one set of options, maybe: `sdk.GRPCSizeOptions`.

OO terminology has infected us all. :)

bzz:
Changed to WIP as I'm looking into CI failure.

common.go
```diff
@@ -1,4 +1,42 @@
...
+	var err error
+	size := maxMessageSizeMb
+	if size >= gRPCMaxMsgSizeCapMb {
+		err = fmt.Errorf("%s=%d is too big (limit is %dMB), using %d instead",
```
creachadair: Do we have a convention about returning errors with non-garbage values? Idiomatically, if err != nil the caller should not use the rest of the response at all. If we do otherwise, we should document that in the function comment. 

I'd suggest we either (a) silently cap the value, and document that we do so; or (b) make the error an actual error.

For (a), 

```
The maximum size limit is capped at X MiB if a larger value is requested.
```

or words to that effect.

common.go
```diff
@@ -1,4 +1,42 @@
...
+			MaxMsgSizeCLIName, size, gRPCMaxMsgSizeCapMb-1, DefaulGRPCMaxMsgSizeMb)
+		size = DefaulGRPCMaxMsgSizeMb
+	}
+
+	size = size * 1024 * 1024
```
creachadair: Please use a separate name, the units have changed:

    sizeBytes := size*(1<<20)

common.go
bzz: Thanks for the heads up!

In cases like this Github has a nice suggestion feature that comes handy - I took liberty of editing your reply directly.

common.go
bzz: Thank you for suggestion - totally agree, but as nothing better crossed my mind I just moved the old name.

```suggestion
func GRPCOptions(maxMessageSizeMb int) ([]grpc.ServerOption, error) {
```

bzz:
Removing WIP as CI is green now. 

Thank you for the feedback @creachadair will address the rest of it tomorrow.

common.go
bzz: Was under impression that "megabits" would be "Mbps" for the thoughtput. 
A [quick](https://github.com/grpc/grpc/commit/61aed92b8262b3a7fa86394cc6b89987ada85228#diff-9d7f390c5994e33fc8848e879931b239R104) [search](https://github.com/grpc/grpc/blob/618a3f561d4a93f263cca23abad086ed8f4d5e86/test/core/compression/message_compress_test.cc#L123) reveals that `MB` seems to be a preferred option here. 👍

common.go
bzz: I see what you mean. But the reason for having this is to get this condition logged, and annoying thing is - logging is only available on the client and we do want to log as this parameter is a frequent source of errors/confusion.

I would leave out propagating a logger (though context or somehow else) for the further refactorings.

So if you prefer strongly only idiomatic use errors, I can revert the logic back and make it un-recoverable condition so the client `os.Exit()` (as it was before).

> If we do otherwise, we should document that in the function comment.
Otherwise, I will go for this option for now.

common.go
bzz: 👍 makes sense, addressed by `MaxSendRecvMsgSizeMB(fs *flag.FlagSet)`

bzz:
Common code moved to `cmd` package, basic test added, functions for creating flags exposed, CI is green now.

Ready for another round.

driver/server/server.go
```diff
@@ -71,6 +73,12 @@ func (s *Server) initialize() error {
 		return err
 	}
 
+	grpcOpts, err := cmdutil.GRPCSizeOptions(*maxMessageSize)
+	if err != nil {
+		s.Logger.Warningf(err.Error())
```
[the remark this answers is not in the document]
juanjux: Plus, it's using grpcOpts with undefined value after the error below so it definitely should exit the function or the program.

common.go
creachadair: I don't think we should conflate separate concerns: Here, logging and error propagation.  Errors have a semantic meaning, while logging is a monitoring/visibility feature. However, I agree that changing the semantics is probably a larger topic than just this PR, and can be addressed separately. So yeah, let's just document it for now.

I'll only note for future reference that lookout already seems to have a context logging package we might re-use: https://godoc.org/github.com/src-d/lookout/util/ctxlog

cmd/util_test.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func TestGRPCOptions_InvalidInput(t *testing.T) {
+	opts, err := GRPCSizeOptions(-1)
+
+	require.Error(t, err, "a given value was not applied")
+	require.NotNil(t, opts)
+	require.Len(t, opts, 2)
+	// does not work as expected for Func values like grpc.ServerOption
+	//require.Contains(t, opts, grpc.MaxRecvMsgSize(DefaulGRPCMaxSendRecvMsgSizeMB))
```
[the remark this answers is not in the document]
bzz: I'm not very keen on turning this into integration test, but still find this comment helpful - next time somebody touches this, he will know the reason why it's not implemented.

But please, let me know if you feel strongly about that - can remove it ofc.

driver/server/server.go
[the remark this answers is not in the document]
bzz: Guys, please, let's keep this discussion in a single place =)

I'm open for reasonable arguments why this approach is better then a current one at https://github.com/bblfsh/bblfshd/pull/216#discussion_r235785584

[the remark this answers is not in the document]
bzz: Agreed to make it an error in https://github.com/bblfsh/bblfshd/pull/216#discussion_r

bzz:
All feedback addressed, @dennwc @juanjux @creachadair, it's ready for another round.

PR #330, opened 2018-11-23
Enable OpenTracing in SDK

dennwc:
Enable OpenTracing and instrument driver operations with spans.

image

### DIFF
--- a/driver/fixtures/fixtures.go
+++ b/driver/fixtures/fixtures.go
@@ -199,6 +199,8 @@ func (s *Suite) testFixturesNative(t *testing.T) {
 }
 
 func (s *Suite) testFixturesUAST(t *testing.T, mode driver.Mode, suf string, blacklist ...string) {
+	ctx := context.Background()
+
 	list, err := ioutil.ReadDir(s.Path)
 	require.NoError(t, err)
 
@@ -226,7 +228,7 @@ func (s *Suite) testFixturesUAST(t *testing.T, mode driver.Mode, suf string, bla
 			name += suffix
 			code := s.readFixturesFile(t, name)
 
-			ctx, cancel := context.WithTimeout(context.Background(), parseTimeout)
+			ctx, cancel := context.WithTimeout(ctx, parseTimeout)
 			ast, err := dr.Parse(ctx, string(code))
 			cancel()
 			if err != nil {
@@ -240,7 +242,7 @@ func (s *Suite) testFixturesUAST(t *testing.T, mode driver.Mode, suf string, bla
 
 			tr := s.Transforms
 			if s.WritePreprocessed {
-				ua, err := tr.Do(driver.ModePreprocessed, code, ast)
+				ua, err := tr.Do(ctx, driver.ModePreprocessed, code, ast)
 				require.NoError(t, err)
 
 				un, err := marshalUAST(ua)
@@ -248,7 +250,7 @@ func (s *Suite) testFixturesUAST(t *testing.T, mode driver.Mode, suf string, bla
 
 				s.writeFixturesFile(t, name+preExt, string(un))
 			}
-			ua, err := tr.Do(mode, code, ast)
+			ua, err := tr.Do(ctx, mode, code, ast)
 			require.NoError(t, err)
 
 			if len(blacklist) != 0 {
@@ -335,6 +337,7 @@ func (s *Suite) benchmarkTransform(b *testing.B, legacy bool) {
 	if s.BenchName == "" {
 		b.SkipNow()
 	}
+	ctx := context.Background()
 	code := s.readFixturesFile(b, s.BenchName+s.Ext)
 
 	tr := s.Transforms
@@ -345,7 +348,7 @@ func (s *Suite) benchmarkTransform(b *testing.B, legacy bool) {
 	require.NoError(b, err)
 	defer dr.Close()
 
-	ctx, cancel := context.WithTimeout(context.Background(), parseTimeout)
+	ctx, cancel := context.WithTimeout(ctx, parseTimeout)
 	rast, err := dr.Parse(ctx, string(code))
 	cancel()
 	if err != nil {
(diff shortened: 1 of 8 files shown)


driver/fixtures/fixtures.go
```diff
@@ -345,7 +348,7 @@ func (s *Suite) benchmarkTransform(b *testing.B, legacy bool) {
 	require.NoError(b, err)
 	defer dr.Close()
 
-	ctx, cancel := context.WithTimeout(context.Background(), parseTimeout)
+	ctx, cancel := context.WithTimeout(ctx, parseTimeout)
```
creachadair: The context isn't used till here, this can more simply be:

    ctx, cancel := context.WithTimeout(context.Background(), parseTimeout)

driver/native/native.go
```diff
@@ -139,8 +141,49 @@ func (r *parseResponse) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+func (d *Driver) writeRequest(ctx context.Context, req *parseRequest) error {
+	deadline, _ := ctx.Deadline()
+
+	sp, _ := opentracing.StartSpanFromContext(ctx, "bblfsh.native.Parse.encodeReq")
+	defer sp.Finish()
+
+	if !deadline.IsZero() {
```
creachadair: I think this'd be a little nicer with a defer:

```
if dl, _ := ctx.Deadline(); !dl.IsZero() {
   d.stdin.SetWriteDeadline(dl)
   defer d.stdin.SetWriteDeadline(time.Time{})
}
```

then you needn't have three separate blocks with semantic connections.

driver/native/native.go
```diff
@@ -139,8 +141,49 @@ func (r *parseResponse) UnmarshalJSON(data []byte) error {
...
+
+	err := d.enc.Encode(req)
+	if !deadline.IsZero() {
+		d.stdin.SetWriteDeadline(time.Time{})
+	}
+	return err
+}
+
+func (d *Driver) readResponse(ctx context.Context) (*parseResponse, error) {
+	deadline, _ := ctx.Deadline()
```
creachadair: As 150, ibid.

driver/server/grpc.go
```diff
@@ -17,46 +18,22 @@ import (
...
-	}()
-
-	return s.Server.Serve(listener)
-}
-
-func (s *GRPCServer) initialize() error {
-	s.Server = grpc.NewServer(s.Options...)
-
-	logrus.Debugf("registering grpc service")
+func NewGRPCServer(drv driver.DriverModule, opts ...grpc.ServerOption) *grpc.Server {
```
creachadair: Is there some reason we're hard-coding these options into the constructor, rather than passing them in from main? (I assume there is a reason, but a comment might be helpful here)

driver/server/server.go
```diff
@@ -27,16 +32,18 @@ var (
...
 // Server is a grpc server for the communication with the driver.
 type Server struct {
-	*GRPCServer
+	grpc *grpc.Server
 	// Logger a logger to be used by the server.
 	Logger Logger
 
 	d driver.DriverModule
+
+	closers []io.Closer
```
creachadair: Please document this field (I realize it's unexported, but the semantics are slightly subtle).

bzz:
Would be nice to rebase on latest master

dennwc:
@bzz @creachadair Done, ready for another round.

driver/server/grpc.go
dennwc: Added a comment

bzz: On API level it makes sense to pass in all grpc options though the constructor, so a client (in this case driver Server) can choose e.g to disable tracing for some production deployment based on CLI.

driver/server/server.go
```diff
@@ -2,16 +2,21 @@ package server
...
+	"google.golang.org/grpc"
+
 	cmdutil "gopkg.in/bblfsh/sdk.v2/cmd"
 	"gopkg.in/bblfsh/sdk.v2/driver"
 	"gopkg.in/src-d/go-errors.v1"
 )
 
 var (
 	ErrInvalidLogger       = errors.NewKind("invalid logger configuration")
+	ErrInvalidTracer       = errors.NewKind("invalid tracer configuration")
```
bzz: I would prefer adding go doc at least for the new code we are adding.

protocol/driver.go
```diff
@@ -43,7 +45,10 @@ type driverServer struct {
 	d driver.Driver
 }
 
-func (s *driverServer) Parse(ctx xcontext.Context, req *ParseRequest) (*ParseResponse, error) {
+func (s *driverServer) Parse(rctx xcontext.Context, req *ParseRequest) (*ParseResponse, error) {
```
bzz: Go doc?

driver/server/grpc.go
dennwc: To disable tracing, you just don't pass any tracing configuration in env :)

Also, tracing does not incur the performance cost you think it does. It may apply to 1/1000 requests, for example.

So I think it's safe to integrate it by default. Because if you forget to enable it, bblfshd will not propagate trace info to the driver, for example.

creachadair: Ok, but it isn't just about cost—this is an API and if the caller is heap profiling (for example), even a 0.1% sampling rate is enough to mess up the results. 

If the only callers of this function are "us" (i.e., clients we maintain), it seem like we could pretty easily make sure they all pass the tracing options (we could even make a slice of them as a variable in the package, or a helper to populate them among other options). If there _are_ other clients, maybe we can separate these two concerns:

```
// NewGRPCServer constructs a GRPC endpoint that dispatches to the specified
// driver module for …
func NewGRPCServer(…) *grpc.Server { … }

// NewTracedGRPCServer constructs a GRPC endpoint with tracing enabled that …
func NewTracedGRPCServer(…) *grpc.Server { … return NewGRPCServer(mod, combinedOpts...) }
```

or words to that effect. I realize it seems like extra work now, but these corner cases pile up.

In any case: We don't have to do all of this right now, but if this constructor is going to have default options baked in, I'd like to see that reflected in the doc comment since it's part of the API for this package.

driver/server/server.go
creachadair: Agreed. E.g., `// ErrInvalidTracer is reported by blah when an invalid trace configuration is found`
or whatever is actually true here.

dennwc:
@creachadair @bzz Done!

dennwc:
Great! The floodgate of PRs to drivers/clients/bblfshd/gitbase is now open :)

PR #333, opened 2018-11-29
Reintroduce InlineImport

dennwc:
`InlineImport` was discussed and approved in the beginning of Semantic UAST conversion, but was disabled in SDK until this time.

Thanks to the great work made by @juanjux on https://github.com/bblfsh/cpp-driver/pull/16, we now need to get this type back to SDK to use in C/C++ driver.

P.S. Sorry, there is no godoc for this type. Should be done separately by dumping parts of the design doc into Go files.

### DIFF
--- a/uast/uast.go
+++ b/uast/uast.go
@@ -40,6 +40,7 @@ func init() {
 		Import{},
 		RuntimeImport{},
 		RuntimeReImport{},
+		InlineImport{},
 		Argument{},
 		FunctionType{},
 		Function{},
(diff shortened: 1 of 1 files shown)


uast/uast.go
```diff
@@ -322,7 +323,7 @@ type RuntimeImport Import
 
 type RuntimeReImport RuntimeImport
 
-//type InlineImport Import
+type InlineImport Import
```
creachadair: Please document this new type.

dennwc: As mentioned in the comment for PR, I'd like to document all types here in a separate PR.

creachadair: SGTM.

PR #334, opened 2018-11-29
DSL validation: unused variables

dennwc:
Currently, the transformation DSL performs the following validations:
* Variables in the destination should be defined in the source.
* All fields of an object should be used, or the transformation should be partial.

This, however, does not include the following validation:
* Variables defined in the source should be used in the destination.

This PR adds test cases for all mentioned validations and implements "unused variable" validation for the DSL.

P.S. Sorry, @juanjux, but you'll need to fix all of them in one way or another in the C++ driver.

### DIFF
--- a/uast/transformer/errors.go
+++ b/uast/transformer/errors.go
@@ -17,6 +17,12 @@ var (
 	// in Lookup or If, for example, before another transformation step that sets this variable is executed.
 	// If this is the case, see Fields vs Obj comparison in regards to execution order.
 	ErrVariableNotDefined = errors.NewKind("variable %q is not defined")
+	// ErrVariableUnused is returned when a first half of transformation defines a variable and the second part
+	// never uses it in any operations.
+	//
+	// If you receive this error and still think that every variable is used - double check conditional branches
+	// like Opt, If, Case, etc.
+	ErrVariableUnused = errors.NewKind("variables %q unused in the second part of the transform")
 	// ErrExpectedObject is returned when transformation expected an object in the tree or variable, but got other type.
 	ErrExpectedObject = errors.NewKind("expected object, got %T")
 	// ErrExpectedList is returned when transformation expected an array in the tree or variable, but got other type.
--- a/uast/transformer/transformer.go
+++ b/uast/transformer/transformer.go
@@ -345,6 +345,9 @@ func (m mappings) Do(root nodes.Node) (nodes.Node, error) {
 		return n, true
 	})
 	err := NewMultiError(errs...)
+	if err == nil {
+		err = st.Validate()
+	}
 	if ok {
 		return nn, err
 	}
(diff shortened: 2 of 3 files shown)


dennwc:
The proposal is not related to the PR itself. As I mentioned in the C++ driver, there are a few ways how to solve it, and you can check Java driver for an example:
https://github.com/bblfsh/java-driver/blob/master/driver/normalizer/normalizer.go#L220
It's not the best thing to do of course, but to make it work properly we need more types in Semantic UAST definition.

Regarding the proposal: what you are proposing is very dangerous and will break the main promise of Semantic UAST.

To give an analogy, image a programming language that has an `extra` keyword. A user can set it to a random value, and the compiler will do something undocumented and random to that value. Would this language be stable and safe to code in?

Same happens here. If we say that "all semantic nodes are exactly the same in each language" and then make an `Extra` backdoor to dump random stuff to it, this will defeat the main purpose of the UAST. Instead of finding the right unified representation we will just lazily dump stuff into `Extra`. This is something that I want to avoid at all cost. Users should trust our UAST. This will break the trust, or will lead to code like `if Extra != nil { // skip }`.

And I totally understand that I'm currently proposing to dump this extra information to another random node in the `FunctionGroup`, but it's only a temporary workaroud that we will remove eventually. All other types work fine without `Extra`. We just need to extend the coverage of SemUAST types to be able to express that information.

dennwc:
Anyway, your concerns are valid and deserve a separate discussion. Let's open a different issue for this. It's not related to this change at all (you can still drop fields with `Any(nil)`).

juanjux:
In the Java driver is almost not ugly (but still ugly-ish) because you have after the `FunctionGroup` node in the `Alias` an object with just three properties:

```
				Obj{ // FIXME: store them as annotations at least
					"constructor":          Var("constr"),
					"thrownExceptionTypes": Var("exc"),
					"typeParameters":       Var("tmpl"),
				},
```

In the case of the C++ driver, since almost all nodes have these qualifiers, I would basically have to add a parallel tree with the same structure and this extra nodes which would be super ugly (and useless to the user). So, for this driver, I think this is a non-solution if we think these properties can't be dropped on the semantic UAST.

I don't think an `Extra` field would break any trust or premises since that field would be documented to hold the properties that doesn't fix in our common-denominator objects and thus they're not really part of the semantic UAST, yet they're easily accessible to the user if he need to check them since it'll be a children of the semantic node. Having the programmer dumping things that he shouldn't there is an error an no different than dumping a property in a separate object if it belonged to the common part, for example.

dennwc:
@juanjux Let's open a separate issue.

If there no other comments to this PR - I will merge it. This change is about unused variables only.

juanjux:
I think not having a non-kludgy solution to unused variables is a problem of this PR since it'll force those. Let's wait for @creachadair and @bzz.

juanjux:
I'm confused now :laughing: 

Either:

1) You think that dropping fields is unacceptable (that was my impression by your sentence in bold in the other PR: **don't drop fields**) because we're losing information and in this case we need to find a nice solution to keep those fields easy to access in case the user want them (which the current ones aren't). This is why I proposed an alternative.

2) Dropping fields is acceptable when we can't realistically map them to the semantic schema, but only explicitly. I think this is not so bad since we defined semantic objects as a common denominator and thus information loss is expected (but the user still have the UAST with all the information).

If its 1) I think we should wait to have (non kludgy) a solution for it before merging this, which we can continue in other place, but still having this depend on that solution. If its 2) it's ok for me to merge this and I'll explicitly drop fields with the `Any(nil)` thing.

In any case, this is an interesting discussion of an important matter (what to do with these fields) and would like for the rest of the team to join. Since I think @bzz is ill, we could make a meeting on monday if you all agree.

juanjux:
Ok, then we agree that this is unacceptable and the current solutions aren't any good either :) so I'll add a meeting at 17.30 PM tuesday (monday is the empathy session) to discuss what to do about this but approve this meanwhile.

PR #337, opened 2018-12-05
Define Ambiguity node

dennwc:
This PR defines a new node type for SemUAST called `Ambiguity`.

This node will represent multiple alternative meanings of the same node in case driver cannot determine its type with a 100% certainty (without running the compiler or symbol resolution), but can propose a small number of alternative meanings.

The first use case is that languages like Python, Go and some others have built-in names like `True`/`False` that can be redefined. The driver might mark them as `Identifier` and be done with it, but this won't be that helpful for the end user. In fact, the driver knows that in this specific language the `True` identifier with 99% certainty is a boolean literal. So instead of dumping a generic `Identifier`, the driver may emit an `Ambiguity` node with both `Bool` and `Identifier` nodes.

The second use case is that sometimes operators might be overloaded in the language itself. I don't want to state that this will solve an issue of user-defined operator overloading, though. The current use case is a bit simpler for now. For example, in Python there is a `%` operator that can be a modulo operator or a string interpolation operator. The driver may emit both possible versions as an `Ambiguity` node if it cannot assert the type of the first operand.

Later, this may help to identify places in UAST that needs additional analysis. And it will immediately allow searching for nodes of more specific types in Babelfish instead of generic ones.

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/337)

### DIFF
--- a/uast/types.go
+++ b/uast/types.go
@@ -212,6 +212,7 @@ func fieldName(f reflect.StructField) (string, bool, error) {
 var (
 	reflString  = reflect.TypeOf("")
 	reflAny     = reflect.TypeOf((*Any)(nil)).Elem()
+	reflAnyArr  = reflect.TypeOf([]Any{})
 	reflNode    = reflect.TypeOf((*nodes.Node)(nil)).Elem()
 	reflNodeExt = reflect.TypeOf((*nodes.External)(nil)).Elem()
 )
@@ -353,6 +354,26 @@ func setAnyOrNode(dst reflect.Value, n nodes.External) (bool, error) {
 			return false, err
 		}
 		dst.Set(reflect.ValueOf(nd))
+		return true, nil
+	} else if rt == reflAnyArr {
+		narr, ok := n.(nodes.ExternalArray)
+		if !ok {
+			return false, nil
+		}
+		sz := narr.Size()
+		arr := make([]Any, 0, sz)
+		for i := 0; i < sz; i++ {
+			e := narr.ValueAt(i)
+			var v Any = e
+			if nv, err := NewValue(TypeOf(e)); err == nil {
+				if err = nodeAs(e, nv); err != nil {
+					return false, err
+				}
+				v = nv.Interface()
+			}
+			arr = append(arr, v)
+		}
+		dst.Set(reflect.ValueOf(arr))
 		return true, nil
 	}
 	return false, nil
--- a/uast/types_test.go
+++ b/uast/types_test.go
@@ -72,6 +72,46 @@ var casesToNode = []struct {
 		},
 	},
 	{
+		name: "Ambiguity",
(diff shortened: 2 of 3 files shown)


creachadair:
> While I think this is a nice feature to have, it was unplanned and the impact of this and the Bool PR is not trivial

If I understand the proposal correctly, this is effectively adding sum types to our abstract type system. That's not necessarily a bad idea—several languages have variations of sum-typing that we might want to incorporate anyway. However, I agree with @juanjux that it's a much larger design question, and we should consider it as a new feature rather than adding it in piecewise.

I tend to agree that short term we should probably be aiming for "breadth" of coverage (e.g., more languages) before "depth" of coverage (e.g., more detailed types). Both are valuable, but I think the immediate impact for our customers is higher if we go broad first.

dennwc:
The aim of the proposal is to start a discussion about this specific problem. Just in case - I'm not proposing to update all the drivers :)

I also agree that extending language coverage should be the highest priority. This PR is more like a brain-dump of the problem I was thinking about in the background. If you think the proposal is reasonable - let's discuss it (same - in the background).

@juanjux Those are really good questions. First, this node type won't break annotation code, because Semantic pass happens before the Annotation pass and the last one is already prepared to deal with arbitrary Semantic types. And annotations won't ever target Semantic types directly, because it doesn't make sense to do so (Semantic types are "roles" themselves).
For the second question, I think you are referring to the normalization pass, not to the annotation pass. There is a way to make this "or" selector work, but I think it's unnecessary. Adding `Ambiguity`, by the definition, will happen as the last transform, and all other transforms will work on a "pure" tree.
And if you are thinking about XPath as well, we can solve it in a better way. Since we control the projection now, and projection knows the schema, we can "squash" the `Ambiguity` node and pretend that there are two alternative nodes in that XML tree. Queries will just work with both types of nodes in this case.

@creachadair I'm not sure that this is similar to sum types. Sum types assume that there is a set of defined types and the sum type may consist of few of them. And this is somehow expressed in AST/code/type system/whatever.
This change, in contrast, aims to describe a specific instance of an AST node and provides alternative interpretations for it. This won't happen automatically for specific node types.

Instead, the transformation code will search for specific patterns (`x % y`, for example, which is `py:BinOp %`) and will try to assert some condition (`x` is a string).
If the assertion is successful, the transform will replace the node with a specific UAST type (string interpolation in this case, if we imagine we define it).
And if the assertion cannot be made (`x` is an identifier, and the driver won't bother with flow analysis to find out the type), the transform will replace the node with an `Ambiguity` (either "modulo" or "string interpolation" node).

I think it's a useful change, in general. For most languages, we know 2-3 possible meanings of operators or some identifiers and we can easily be more specific than "unknown binop %" or "identifier `true`".

PR #338, opened 2018-12-05
Define Bool node (boolean literal)

dennwc:
We already have a `String` literal node, and this PR goes one step further and defines a boolean literal node (`Bool`) in SemUAST.

The definition of the node is pretty streightforward: it's a boolean value that can be either "true" or "false".

This node type is useless for some drivers without https://github.com/bblfsh/sdk/pull/337.

### DIFF
--- a/uast/types_test.go
+++ b/uast/types_test.go
@@ -72,6 +72,27 @@ var casesToNode = []struct {
 		},
 	},
 	{
+		name: "Bool",
+		obj: Bool{
+			GenNode: GenNode{
+				Positions: Positions{
+					KeyStart: {Offset: 3, Line: 2, Col: 1},
+					KeyEnd:   {Offset: 8, Line: 2, Col: 6},
+				},
+			},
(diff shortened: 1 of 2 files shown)


juanjux:
I approved, since this is worthy and simple, but my approval doesn't imply we thing we should upgrade all the drivers to this immediately (see my comment here:  https://github.com/bblfsh/sdk/pull/337#issuecomment-444805728).

dennwc:
Right. I don't think we should update all drivers immediately. I will probably update 2-3 of them when I have some time just to make sure it works.

@bzz Correct, this is a part of a bigger change, but this change may span a few years. We can gradually improve the coverage of SemUAST types when it makes sense.

For `Bool` the definition is clear, but for other things, it won't be that easy. So right now I identified few common nodes across drivers that are low-hanging fruits and will try to define a schema for them to keep things going. But we have other more important tasks, of course.

PR #341, opened 2019-01-03
Improve documentation (part 1)

dennwc:
Improve documentation for a `uast` package. Covers the UAST schema and related functions.

### DIFF
--- a/uast/types.go
+++ b/uast/types.go
@@ -10,7 +10,9 @@ import (
 )
 
 var (
-	ErrIncorrectType     = errors.NewKind("incorrect object type: %q, expected: %q")
+	// ErrIncorrectType is returned when trying to load a generic UAST node into Go value of an incorrect type.
+	ErrIncorrectType = errors.NewKind("incorrect object type: %q, expected: %q")
+	// ErrTypeNotRegistered is returned when trying to create a UAST type that was not associated with any Go type.
 	ErrTypeNotRegistered = errors.NewKind("type is not registered: %q")
 )
 
@@ -46,6 +48,16 @@ func (n nodeID) String() string {
 	return n.NS + ":" + n.Name
 }
 
+// RegisterPackage registers UAST namespace and associates listed types with it.
+//
+// Example:
+//   type Node struct{}
+//
+//   func init(){
+//      // will register a UAST type "my:Node" associated with
+//      // a Node type from this package
+//      RegisterPackage("my", Node{})
+//   }
 func RegisterPackage(ns string, types ...interface{}) {
 	if _, ok := namespaces[ns]; ok {
 		panic("namespace already registered")
@@ -73,6 +85,9 @@ func RegisterPackage(ns string, types ...interface{}) {
 	}
 }
 
+// LookupType finds a Go type corresponding to a specified UAST type.
+//
+// It only returns types registered via RegisterPackage.
 func LookupType(typ string) (reflect.Type, bool) {
 	name := parseNodeID(typ)
 	rt, ok := name2type[name]
@@ -139,6 +154,9 @@ func NewObjectByTypeOpt(typ string) (obj, opt nodes.Object) {
 	return obj, opt
 }
 
+// NewValue creates a new Go value corresponding to a specified UAST type.
+//
+// It only creates types registered via RegisterPackage.
 func NewValue(typ string) (reflect.Value, error) {
 	rt, ok := LookupType(typ)
 	if !ok {
@@ -152,6 +170,11 @@ func NewValue(typ string) (reflect.Value, error) {
 	}
 }
 
+// TypeOf returns the UAST type of a value.
+//
+// If the value is a generic UAST node, the function return the value of KeyType.
+//
+// If an object is registered as a UAST schema type, the function returns the associated type.
 func TypeOf(o interface{}) string {
 	switch obj := o.(type) {
 	case nil:
@@ -216,8 +239,8 @@ var (
 	reflNodeExt = reflect.TypeOf((*nodes.External)(nil)).Elem()
 )
 
-// ToNode converts objects returned by schema-less encodings such as JSON to Node objects.
-// It also supports types from packages registered via RegisterPackage.
+// ToNode converts generic values returned by schema-less encodings such as JSON to Node objects.
+// It also supports values registered via RegisterPackage.
 func ToNode(o interface{}) (nodes.Node, error) {
 	return nodes.ToNode(o, toNodeFallback)
 }
@@ -332,6 +355,9 @@ func structToNode(obj nodes.Object, rv reflect.Value, rt reflect.Type) error {
 	return nil
 }
 
+// NodeAs loads a generic UAST node into provided Go value.
+//
+// It returns ErrIncorrectType in case of type mismatch.
 func NodeAs(n nodes.External, dst interface{}) error {
 	var rv reflect.Value
 	if v, ok := dst.(reflect.Value); ok {
--- a/uast/types.go
+++ b/uast/types.go
@@ -10,9 +10,11 @@ import (
 )
 
 var (
-	// ErrIncorrectType is returned when trying to load a generic UAST node into Go value of an incorrect type.
+	// ErrIncorrectType is returned when trying to load a generic UAST node into a Go value
+	// of an incorrect type.
 	ErrIncorrectType = errors.NewKind("incorrect object type: %q, expected: %q")
-	// ErrTypeNotRegistered is returned when trying to create a UAST type that was not associated with any Go type.
+	// ErrTypeNotRegistered is returned when trying to create a UAST type that was not associated
+	// with any Go type. See RegisterPackage.
 	ErrTypeNotRegistered = errors.NewKind("type is not registered: %q")
 )
 
@@ -48,7 +50,9 @@ func (n nodeID) String() string {
 	return n.NS + ":" + n.Name
 }
 
-// RegisterPackage registers UAST namespace and associates listed types with it.
+// RegisterPackage registers a new UAST namespace and associates the concrete types
+// of the specified values with it. All types should be in the same Go package.
+// The name of each type is derived from its reflect.Type name.
 //
 // Example:
 //   type Node struct{}
@@ -172,7 +176,7 @@ func NewValue(typ string) (reflect.Value, error) {
 
 // TypeOf returns the UAST type of a value.
 //
-// If the value is a generic UAST node, the function return the value of KeyType.
+// If the value is a generic UAST node, the function returns the value of its KeyType.
 //
 // If an object is registered as a UAST schema type, the function returns the associated type.
 func TypeOf(o interface{}) string {
--- a/uast/uast.go
+++ b/uast/uast.go
@@ -49,10 +49,10 @@ func init() {
 
 // Special field keys for nodes.Object
 const (
-	KeyType  = "@type"  // InternalType
-	KeyToken = "@token" // Token
-	KeyRoles = "@role"  // Roles, for representations see RoleList
-	KeyPos   = "@pos"   // All positional information is stored in this field
+	KeyType  = "@type"  // the type of UAST node (InternalType in v1)
+	KeyToken = "@token" // token of the UAST node (Native and Annotated nodes only)
+	KeyRoles = "@role"  // roles of UAST node (Annotated nodes only); for representations see RoleList
+	KeyPos   = "@pos"   // positional information is stored in this field, see Positions
 )
 
 const (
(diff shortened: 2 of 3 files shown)


uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
 	GenNode
+	// Names is a list of simple identifiers starting from a root level of hierarchy
+	// and ending with leaf identifier.
 	Names []Identifier `json:"Names"`
 }
 
+// Comment is a no-op node that can span multiple lines and provides a human-readable
+// description for code around it.
+//
+// TODO: currently some annotations are also considered a Comment; need to clarify this
```
creachadair: Can you please include an issue ref? E.g., `TODO(#123): …`

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
 type Comment struct {
 	GenNode
-	Text   string `json:"Text"`
+
+	// Block is set to true for block-style comments.
```
creachadair: What distinguishes a block-style comment in terms of this representation? Both line- and block-comments have text—is the distinction that block:true means there may be line breaks?

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
+	//
+	// TODO: should be a string similar to Format field in String literal;
+	//       may have more than 2 possible values (line, block, doc?)
+	Block bool `json:"Block"`
+
+	// Text is an unescaped UTF8 string with the comment text.
+	//
+	// Drivers should trim any comment-related tokens as well as whitespaces and
```
creachadair: s/should/must/ ?

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
+	//
+	// Example:
+	//    /*
+	//     * some comment
+	//     */
+	//
+	//    only "some comment" is considered a text
+	Text string `json:"Text"`
+
+	// Prefix is usually a set of whitespaces that appear before the first line of an
```
creachadair: What if it's not consistent? E.g.,

```
/*
 * Some comment:
 *     - Hello
 *     - Goodbye
 * some more comment.
 */
```
The actual first line has `""`, the first interesting line has `" "`, and so on.

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
+	//
+	// Example:
+	//    /*
+	//     * some comment
+	//     */
+	//
+	//    the "\n" after the "/*" token is considered a prefix
 	Prefix string `json:"Prefix"`
+
+	// Suffix is usually a set of whitespaces that appear after the last line of an
```
creachadair: "Usually"?

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
+	// A UAST node to assign a name to.
+	Node Any `json:"Node"`
+
+	// TODO: should include a pointer to a scope where an alias is defined
 	// Target *Scope
 }
 
+// Import is a statement that can load other modules into the program or library.
+//
+// This is a declarative import statement. It's position in the UAST does not affect
```
creachadair:
```suggestion
 // This is a declarative import statement. Its position in the UAST does not affect
```

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
+//
+// Similar to import in PHP, Python.
+//
+// For other import types, see Import.
 type RuntimeReImport RuntimeImport
 
+// InlineImport is a subset of import statement that acts like a preprocessor - all statements in
+// the imported module are copied into a position of the UAST node.
+//
+// Similar to import in C and C++.
```
creachadair:
```suggestion
 // Similar to #include in C and C++.
```

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
 	// Target *Scope
 }
 
+// Import is a statement that can load other modules into the program or library.
+//
+// This is a declarative import statement. It's position in the UAST does not affect
+// the way and the time when the module is imported and the side-effects are executed
+// only once a package is initialized.
+//
+// Similar to import in Go, Java, C#.
```
creachadair: Rather than "similar to…" in these examples, I'd suggest wording it to more directly describe the relationship. Perhaps: "This describes imports in Go, Java, and C#, for example."

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
-	Variadic    bool        `json:"Variadic"`
-	MapVariadic bool        `json:"MapVariadic"`
-	Receiver    bool        `json:"Receiver"`
+	// Name is an optional name of an argument.
+	Name *Identifier `json:"Name"`
+
+	// Type is an optional type of an argument.
+	Type Any `json:"Type"`
+
+	// Init is an expression that should be used to initialize an argument
```
creachadair:
```suggestion
 	// Init is an optional expression used to initialize the argument
```

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
+	// in case no value is provided. Also known as a default value of an argument.
```
creachadair: Subtly, the default value is what the expression yields, whereas `Init` is the expression itself.

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
+	Init Any `json:"Init"`
+
+	// Variadic flag may be set only for the last argument to denote a function
```
creachadair:
```suggestion
 	// Variadic is set true for the last argument of a function
```

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
+	// with a variadic number of arguments.
+	Variadic bool `json:"Variadic"`
+
+	// MapVariadic flag may be set only for the last argument to denote a function
```
creachadair:
```suggestion
 	// MapVariadic flag is set true for the last argument of a function
```

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
+	// Receiver is set to true if an argument is a receiver of a method call.
+	Receiver bool `json:"Receiver"`
 }
 
+// FunctionType is a signature of a function.
 type FunctionType struct {
 	GenNode
+	// Arguments is a set of arguments the function accepts.
+	//
+	// Methods defined on structures and classes should have the first argument
```
creachadair: "must" ?

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
+	// that corresponds to a method's receiver ("this" in most languages).
 	Arguments []Argument `json:"Arguments"`
-	Returns   []Argument `json:"Returns"`
+
+	// Returns is a set of values returned by a function.
+	//
+	// Languages with an implicit return should specify a single return with an
```
creachadair: Does this imply that if `len(a.Returns) == 0` that it's explicitly a procedure returning no values?

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
+// What is considered a Function:
+// - function declaration;
+// - anonymous functions;
 type Function struct {
 	GenNode
+	// Type is a signature of a function. Should always be set.
 	Type FunctionType `json:"Type"`
-	Body *Block       `json:"Body"`
+
+	// Body is an optional implementation of a function. should point to a Block with
```
creachadair: Why is this optional? If the declaration has no body, shouldn't it be a `FunctionType`?

dennwc:
You may have just commented "wrap at 90" instead of wasting time writing those changes on the Github :) Reconfigured the editor (it was 120).

uast/types.go
```diff
@@ -46,6 +48,16 @@ func (n nodeID) String() string {
...
+//
+// Example:
+//   type Node struct{}
+//
+//   func init(){
+//      // will register a UAST type "my:Node" associated with
+//      // a Node type from this package
+//      RegisterPackage("my", Node{})
+//   }
 func RegisterPackage(ns string, types ...interface{}) {
```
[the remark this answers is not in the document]
dennwc: Clarified in the comment - it requires all types to be in the same package.

uast/uast.go
```diff
@@ -87,18 +87,26 @@ type Position struct {
...
 }
 
+// Valid checks if position value is valid.
 func (p Position) Valid() bool {
 	return p != (Position{})
 }
 
+// Less returns true if the current position is less than an argument.
+//
+// If both positions has an offset, it will be used for comparison. Otherwise, line-column pair will be used.
```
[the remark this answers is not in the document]
dennwc: Yes, line/col can get out of sync. But we prevent that by recalculating either offset->line/col or line/col->offset depending on what is easier for the driver.

uast/uast.go
```diff
@@ -117,7 +125,7 @@ func (p Position) Less(p2 Position) bool {
 	return p.Col != 0 && p.Col < p2.Col
 }
 
-// Positions is a container object that stores all positional information for a node.
+// Positions is a container object that stores all positional information for a UAST node.
```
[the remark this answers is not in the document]
dennwc: Clarified - the key is a name of position info. There are two standard names for start/end and other names can be included in the driver supports it.

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
 // Scope is a temporary definition of a scope semantic type.
 type Scope = Any
 
+// GenNode is a base of every UAST node. It stores positional information.
```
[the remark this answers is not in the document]
dennwc: Well, maybe. But there might be other common fields added later. Don't want to copy-paste it to each node type.

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
 type GenNode struct {
 	Positions Positions `json:"@pos,omitempty"`
 }
 
+// Identifier is a name of an entity.
```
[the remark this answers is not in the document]
dennwc: Go comments should start with the name of the type.

uast/uast.go
```diff
@@ -262,87 +270,311 @@ type Any interface{}
...
 }
 
+// QualifiedIdentifier is a name of an entity that consists of multiple simple identifiers,
+// organized in a hierarchy, similar to filesystem paths.
+//
+// What is considered a QualifiedIdentifier:
+// - qualified names that consist of Identifier-like elements;
+//
+// Not considered a QualifiedIdentifier:
+// - path-like or url-like package names (see String);
```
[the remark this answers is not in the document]
dennwc: As far as I know, Java has a pretty direct form of `QualifiedIdentifiers`. So Java's hierarchy is mapped directly to this type. I don't think Java needs `String` representation of it. Go on the other hand does. But maybe you are right and this distinction is a bit artificial.

uast/uast.go
dennwc: It doesn't have an issue associated with it. I need to check which driver does it.

uast/uast.go
dennwc: While writing those comments I also thought that this may be useless for the end user. `Format` should be more useful. Or we can leave it in a token stream if we decide to go that way.

uast/uast.go
dennwc: Your example is still pretty consistent.

The first line has "\n" which qualifies as a Prefix. Or other lines won't include it.
Following lines all have " * " as a Tab. Still consistent.
And the last line has "\n " as a Suffix. Again, Tab does not apply here and Suffix does not affect other lines.

It seems like the naming is bad though. I guess the Suffix appears to be a suffix of each line instead of the whole comment. Is that what you meant by this concern?

uast/uast.go
dennwc: Right. I will drop the last sentence.

uast/uast.go
dennwc: I assumed forward declaration may clarify as `Function` as well as a function header that is implemented in another language (Go function with WASM implementation, Java and JNI).

uast/types.go
```diff
@@ -152,6 +174,11 @@ func NewValue(typ string) (reflect.Value, error) {
 	}
 }
 
+// TypeOf returns the UAST type of a value.
+//
+// If the value is a generic UAST node, the function returns the value of its KeyType.
+//
+// If an object is registered as a UAST schema type, the function returns the associated type.
```
bzz: Just to help me understand the convention/consistency - when an empty newline shall be added to multiple comments and when not?

dennwc: If I remember correctly, godoc merges lines into a single paragraph if there is no newline in between.

uast/uast.go
dennwc: The second case won't have Tab set at all, since, as you mentioned, it's not consistent. And there is no minimal prefix for each line that matches all the lines.

Currently, we have a function in `uast/transform` that does the splitting for all the drivers that request it. Although it's not implemented yet to a full degree (Tab is not set).

PR #342, opened 2019-01-09
positioner: simplify, fix and add more tests

dennwc:
As a part of an effort to fix positioning information in some drivers, this PR revisits the test cases for `positioner` object that fills in this information for drivers.

### DIFF
--- a/uast/transformer/positioner/positions_test.go
+++ b/uast/transformer/positioner/positions_test.go
@@ -108,6 +108,8 @@ func TestFillOffsetEmptyFile(t *testing.T) {
 }
 
 func TestPosIndex(t *testing.T) {
+	// Verify that a multi-byte Unicode rune does not displace offsets after
+	// its occurrence in the input. Test few other simple cases as well.
 	const source = `line1
 ё2
 a3`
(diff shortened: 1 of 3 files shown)


uast/transformer/positioner/positions_test.go
```diff
@@ -65,23 +86,56 @@ func TestFillOffsetFromLineCol(t *testing.T) {
...
 	}
 
 	p := NewFillOffsetFromLineCol()
 	out, err := p.OnCode(data).Do(input)
 	require.NoError(err)
 	require.Equal(expected, out)
 }
+
+func TestPosIndex(t *testing.T) {
+	const source = `line1
```
creachadair: Please add a comment as to what's important about this example. Perhaps:

```
// Verify that a multi-byte Unicode rune does not displace offsets after
// its occurrence in the input.
```

or words to that effect.

uast/uast.go
```diff
@@ -82,7 +82,7 @@ type Position struct {
 	Offset uint32 `json:"offset"`
 	// Line is the line number. It is a 1-based index.
 	Line uint32 `json:"line"`
-	// Col is the column number (the byte offset of the position relative to
+	// Col is the column number - the byte offset of the position relative to
```
creachadair: Is the "-" here meant to be a dash or a minus?

If a dash, please use either "—" or "--" to make it clear (since if it's subtraction, the description has become wrong I think).

PR #343, opened 2019-01-14
Support Unicode character offsets

dennwc:
Currently, SDK only supports positional offset in bytes or line-column pair, where a column is given in bytes.

PR implements a new `Positioner` that converts Unicode character offsets to a byte offset and a line-column pair (in bytes as well).

Also, refactor long name for position functions to something shorter and place a deprecation note on the old ones.

This is change will allow to solve issues like https://github.com/bblfsh/javascript-driver/issues/25.

### DIFF
--- a/uast/transformer/positioner/positions.go
+++ b/uast/transformer/positioner/positions.go
@@ -55,7 +55,12 @@ type Positioner struct {
 
 // OnCode uses the source code to update positional information of UAST nodes.
 func (t Positioner) OnCode(code string) transformer.Transformer {
-	idx := newPositionIndex([]byte(code), t.unicode)
+	var idx *positionIndex
+	if t.unicode {
+		idx = newPositionIndexUnicode([]byte(code))
+	} else {
+		idx = newPositionIndex([]byte(code))
+	}
 	return transformer.TransformObjFunc(func(o nodes.Object) (nodes.Object, bool, error) {
 		pos := uast.AsPosition(o)
 		if pos == nil {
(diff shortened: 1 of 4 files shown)


juanjux:
Nice!

uast/transformer/positioner/positions.go
```diff
@@ -70,19 +93,77 @@ func fillLineColFromOffset(idx *positionIndex, pos *uast.Position) error {
...
+}
+
 type positionIndex struct {
 	offsetByLine []int
+	runeSpans    []runeSpan
 	size         int
 }
 
-func newPositionIndex(data []byte) *positionIndex {
+func newPositionIndex(data []byte, unicode bool) *positionIndex {
```
creachadair: Rather than a Boolean flag, can this be two functions? The shared parts are smaller than the switched part, and I think it'd be better to just give them different names at that point.

uast/transformer/positioner/positions.go
```diff
@@ -70,19 +93,77 @@ func fillLineColFromOffset(idx *positionIndex, pos *uast.Position) error {
...
+	if err != nil {
+		return err
+	}
+	pos.Offset = uint32(off)
+	return fromOffset(idx, pos)
+}
+
+// runeSpan represents a sequence of UTF8 characters of the same size in bytes.
+type runeSpan struct {
+	runeOff  int // index of the first rune
```
creachadair: If this is an index, the name probably shouldn't imply "offset". Maybe: `firstRuneIndex`? I assume this is meant to indicate the _n_th rune from the beginning of the input?

uast/transformer/positioner/positions.go
```diff
@@ -70,19 +93,77 @@ func fillLineColFromOffset(idx *positionIndex, pos *uast.Position) error {
...
+	byteOff  int // bytes offset of the first rune
+	runes    int // number of runes
```
creachadair: Suggestion: `numRunes` or `runeCount`

uast/transformer/positioner/positions.go
```diff
@@ -70,19 +93,77 @@ func fillLineColFromOffset(idx *positionIndex, pos *uast.Position) error {
...
+	runeSize int // in bytes
+}
+
 type positionIndex struct {
 	offsetByLine []int
+	runeSpans    []runeSpan
```
creachadair: Could probably be simply `spans`, since the rune-ness is a property of the value not the field.

dennwc:
@creachadair PTAL

PR #344, opened 2019-01-14
uast: properly load node arrays with registered types

dennwc:
Some UAST schema types may contain fields with arrays of nodes. This change fixes the problem with loading such structs from generic UAST nodes to concrete Go types (currently it just fails).

### DIFF
--- a/uast/types.go
+++ b/uast/types.go
@@ -237,10 +237,11 @@ func fieldName(f reflect.StructField) (string, bool, error) {
 }
 
 var (
-	reflString  = reflect.TypeOf("")
-	reflAny     = reflect.TypeOf((*Any)(nil)).Elem()
-	reflNode    = reflect.TypeOf((*nodes.Node)(nil)).Elem()
-	reflNodeExt = reflect.TypeOf((*nodes.External)(nil)).Elem()
+	reflString   = reflect.TypeOf("")
+	reflAny      = reflect.TypeOf((*Any)(nil)).Elem()
+	reflAnySlice = reflect.TypeOf([]Any{})
+	reflNode     = reflect.TypeOf((*nodes.Node)(nil)).Elem()
+	reflNodeExt  = reflect.TypeOf((*nodes.External)(nil)).Elem()
 )
 
 // ToNode converts generic values returned by schema-less encodings such as JSON to Node objects.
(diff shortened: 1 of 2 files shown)


dennwc:
I believe it has no use right now, it's a backport from https://github.com/bblfsh/sdk/pull/337.

But, if we will have any UAST schema similar to this:
```golang
type MyNode struct{
  List []Any
}
```
the SDK will fail to convert generic UAST nodes like
```golang
nodes.Object{
  "@type": "uast:MyNode",
  "List": Arr{...}
}
```
to a Go struct of type `MyNode`.

This is a bug in the current reflection code in SDK.

juanjux:
LGTM, thanks for the explanation!

juanjux:
Nice to have: some test to check we don't have regressions (and shut up codecov).

uast/types.go
```diff
@@ -239,6 +239,7 @@ func fieldName(f reflect.StructField) (string, bool, error) {
 var (
 	reflString  = reflect.TypeOf("")
 	reflAny     = reflect.TypeOf((*Any)(nil)).Elem()
+	reflAnyArr  = reflect.TypeOf([]Any{})
```
creachadair: Nit: This is a slice, not an array. The difference is significant since an array type and a slice type do not reflect to the same mirror. I suggest: `reflAnySlice`.

uast/types.go
```diff
@@ -384,6 +385,26 @@ func setAnyOrNode(dst reflect.Value, n nodes.External) (bool, error) {
 		}
 		dst.Set(reflect.ValueOf(nd))
 		return true, nil
+	} else if rt == reflAnyArr {
+		narr, ok := n.(nodes.ExternalArray)
+		if !ok {
+			return false, nil
+		}
+		sz := narr.Size()
+		arr := make([]Any, 0, sz)
```
creachadair: Since you know the exact size anyway, and there's no early exit, you might as well just allocate and assign rather than appending: `make([]Any, sz)` … `arr[i] = v`

dennwc: I usually prefer not to. Because an exact size is only an optimization, and shouldn't affect the code below.

uast/types.go
dennwc: Right, arrays are used so rarely that I usually neglect the difference when it comes to naming. But yes, it matters here because of the reflection.

dennwc:
Stolen an original test case from Ambiguity PR (this change is a backport from it). Should be good to go.

uast/types.go
creachadair: Generally if you don't care about optimality, I recommend the simpler code: `var arr []Any`. The `append` builtin reallocates intelligently to avoid quadratic moves. If the allocation really matters, it's better not to append at all, and assign directly. The only time this pattern really makes sense is if you're copying out of a map, where there is no index without a separate explicit induction variable.

dennwc: It's more like a middle ground. When I don't care about optimality, I use `var arr []Any`, but when I do care, I simply change that single line to `arr := make([]Any, 0, sz)` and don't touch the code below (it's valid for both cases).

You may argue that it's might be less optimal than a direct assignment, since `append` will increment the counter, but it should be negligible comparing to any allocation and the code is a bit more flexible in case some decide to add additional items. I think it's a good balance, but maybe I should use the optimal one, as you suggest.

PR #345, opened 2019-01-14
More DSL operations to check objects

dennwc:
This is a combination of a few smaller changes required for C# driver.

First, it contains a change that breaks backward compatibility of one rarely used op for the sake of better naming. Previously `Any(sel)` operator was used to check if "any node in the array matches `sel`". PR renames this operator to `AnyElem` and defines a new function `Any()` that acts like "any node". This will allow us to replace a common pattern of `AnyNode(nil)` with `Any()`, which should be more readable.

The change won't break any drivers until they decide to update SDK version. I propose to just bump a minor version since the operation is very rare and the fix will only require to change the function name.

Next, PR defines a `HasFields` helper that can be used to check for presence or absence of specific object fields. Instead of `Not(Has{k1: Any()})` it's now possible to write `HasFields{k1: false}`.

To make the new operator useful, PR also defines a `CheckObj` helper that will properly propagate object-related information to the parent.

It also adds or clarifies documentation for related types.

The change may improve the performance of `AnnotateIfNoRoles` since it will be now able to propagate constraints to the root of the transform tree.

### DIFF
--- a/uast/transformer/arrays.go
+++ b/uast/transformer/arrays.go
@@ -67,6 +67,63 @@ func One(op Op) ArrayOp {
 	return Arr(op)
 }
 
+// AnyElem check matches if any of list elements matches sub-check.
+func AnyElem(s Sel) Sel {
+	if s == nil {
+		s = Is(nil)
+	}
+	return &opAnyElem{sel: s}
+}
+
+type opAnyElem struct {
+	sel Sel
+}
+
+func (*opAnyElem) Kinds() nodes.Kind {
+	return nodes.KindArray
+}
+
+func (op *opAnyElem) Check(st *State, n nodes.Node) (bool, error) {
+	l, ok := n.(nodes.Array)
+	if !ok {
+		return false, nil
+	}
+	for _, o := range l {
+		if ok, err := op.sel.Check(st.Clone(), o); err != nil {
+			return false, err
+		} else if ok {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+// All check matches if all list elements matches sub-check.
+func All(s Sel) Sel {
+	return &opAll{sel: s}
+}
+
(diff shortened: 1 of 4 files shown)


juanjux:
Nothing like writing a new driver to discover possible usability improvements!

Loving these changes but you know what I'm going to say... need some tests :)

dennwc:
@creachadair In general, I agree that those breaking changes should not be accepted lightly. But in this particular case, it's really internal detail of an SDK (no 3rd-party drivers) and only Java driver uses `Any` (I've checked it). So I think it's safe to apply the change until it's too late and update that single driver (it's <30 mins of work).

creachadair:
> @creachadair In general, I agree that those breaking changes should not be accepted lightly. But in this particular case, it's really internal detail of an SDK (no 3rd-party drivers) and only Java driver uses `Any` (I've checked it). So I think it's safe to apply the change until it's too late and update that single driver (it's <30 mins of work).

That's reasonable. I would just like to minimize the amount of state someone (and in this case I'm selfishly thinking of myself) has to keep in mind when making changes to other parts of the system, so that it will be easier to work in parallel.

juanjux:
I'm ok with merging this if we inmediately (in the sense of not waiting for the next SDK update to avoid surprises and maybe find possible bugs now) fix all the drivers needing or benefitting from it so our code remains "idiomatic". AFAIK, this would be the Java driver (incompatible changes) and the C++ driver (replacing `AnyNode(nil)` for `Any()`).

dennwc:
Sounds good. I can update both.

dennwc:
Added a test case. PTAL.

dennwc:
As a bonus, I've added the test for `CaseObj` and documented it. Codecov is finally happy :)

PR #346, opened 2019-01-17
Support token extraction

dennwc:
Changes in this PR enable the token extraction based on the source file and positional information. Needed for C# driver. And it can be used to improve other drivers as well.

- Implement a new `CodeTransformer` that uses positional offsets to update the token of AST nodes.
- Allow checking that positional information matches the token during fixtures tests.
- Add a new code-assisted transformation stage after `Preprocess` stage.
- Document the AST transformation pipeline.

This change also adds a deprecation note to the `Code` transformation stage. It was documented to run on the raw AST (`Native` stage), but it was always running at the end of the pipeline after all annotations were executed.

Drivers rely on real behavior, but there is no reason to run this stage so late in the pipeline. In fact, this was the only stage that has access to the source code. And because it runs so late, it's useless for fixing tokens and positional information - annotation will observe broken or empty tokens. For example, `Normalize` stage (aka Semantic mode) will need correct tokens to unescape them, and it's not possible with the late `Code` stage.

Instead, the new `PreprocessCode` stage was defined. It will run after `Preprocess` stage to be able to see the fixes in the native AST, as well as normalized positional info. And the stage will run before any annotations, so they will be able to see correct positions and/or tokens in the AST.

### DIFF
--- a/driver/transforms.go
+++ b/driver/transforms.go
@@ -9,26 +9,69 @@ import (
 	"gopkg.in/bblfsh/sdk.v2/uast/transformer"
 )
 
-// Transforms describes a set of AST transformations this driver requires.
+// Transforms describes a set of AST transformations the driver requires.
+//
+// The pipeline can be illustrated as:
+//         ( AST )--------------> ( ModeNative )
+//            V
+//      [ Preprocess ]
+//    [ PreprocessCode ]--------> ( ModePreprocessed )
+//            |
+//            |----------------\
+//            |           [ Normalize ]
+//     [ Annotations ] <-------/
+//            |
+//            V
+//        [ Code ]-------------\
+//            |           [ Namespace ]
+//            |                |
+//            V                V
+//     ( ModeAnnotated ) ( ModeSemantic )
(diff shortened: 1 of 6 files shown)


driver/transforms.go
```diff
@@ -9,26 +9,69 @@ import (
...
-	// Preprocess transforms normalizes native AST.
-	// It usually includes:
-	//	* changing type key to uast.KeyType
-	//	* changing token key to uast.KeyToken
-	//	* restructure positional information
+
+	// Preprocess stage normalizes native AST for both Annotated and Semantic stages.
+	//
+	// It usually:
+	//	* changes type key to uast.KeyType
```
creachadair:
```suggestion
	//  * changes type key to uast.KeyType
```

driver/transforms.go
```diff
@@ -9,26 +9,69 @@ import (
...
+	//	* restructures positional information
```
creachadair:
```suggestion
	//  * restructures positional information
```

PR #348, opened 2019-01-18
Allow to use PreprocessCode stage

dennwc:
During the last update of SDK I forgot to update managed skeleton files for drivers. Because of this, it's impossible to use the new `PreprocessCode` stage introduced in the new version (see CI failure in https://github.com/bblfsh/csharp-driver/pull/14).

PR updates skeleton files to allow the new stage.

This, however, will have some consequences for other drivers. Once an SDK is updated, it will include a new version of managed `transforms.go` file, that references a `PreprocessCode` global, which won't be available in old drivers. The fix is simple - include an empty stage with this name.

Another option may be to remove `transforms.go` from managed files. But it was added to this list on purpose. Each time SDK changes the transformation pipeline, the change should be considered by the driver's author. So I don't consider this a good option.

### DIFF
--- a/etc/skeleton/driver/normalizer/normalizer.go
+++ b/etc/skeleton/driver/normalizer/normalizer.go
@@ -32,5 +32,9 @@ var Preprocessors = []Mapping{
 	}.Mapping(),
 }
 
+// PreprocessCode is a preprocessor stage that can use the source code to
+// fix tokens and positional information.
+var PreprocessCode = []CodeTransformer{}
+
 // Normalizers is the main block of normalization rules to convert native AST to semantic UAST.
 var Normalizers = []Mapping{}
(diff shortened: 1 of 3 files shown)


etc/skeleton/driver/normalizer/normalizer.go
```diff
@@ -32,5 +32,9 @@ var Preprocessors = []Mapping{
 	}.Mapping(),
 }
 
+// PreprocessCode is a preprocessor stage that can use the source code to
+// fix tokens and positional information.
+var PreprocessCode = []CodeTransformer{}
```
creachadair: Why a type alias instead of a type?

creachadair: I am apparently out of it. :)

PR #350, opened 2019-01-25
Recover from read timeouts in native driver

dennwc:
Currently, there might be a case when the client sends a request to the native driver, and then cancels it and returns with a timeout error.

But there is no way to cancel the request on this simple native protocol that we have. The native driver will keep processing the request and will send a response at some point. When the second client will try to send another request, he might receive this first "lagged" response instead of the right one.

This PR detects timeout error in the native driver server (Go) and sets a flag to discard an old response before attempting to send the next request.

Also, improve tests for the native driver.


driver/native/native.go
```diff
@@ -142,33 +152,58 @@ func (r *parseResponse) UnmarshalJSON(data []byte) error {
...
+		panic(fmt.Errorf("unexpected state: %v", d.state))
+	}
+	sp, _ := opentracing.StartSpanFromContext(ctx, "bblfsh.native.Parse.skipResp")
 	defer sp.Finish()
 
-	if !deadline.IsZero() {
-		d.stdout.SetReadDeadline(deadline)
-		defer d.stdout.SetReadDeadline(time.Time{})
+	// TODO: ideally we should just skip it without decoding
+	var r struct{}
```
creachadair: You can avoid most of the work of decoding by passing `var r json.RawMessage`. It will still have to parse the message, but will not construct memory objects for it apart from the byte array.

driver/native/native.go
```diff
@@ -191,6 +226,26 @@ func (d *Driver) Parse(rctx context.Context, src string) (nodes.Node, error) {
 	d.mu.Lock()
 	defer d.mu.Unlock()
 
+	if d.state != stateOK && d.state != stateTimeout {
+		return nil, driver.ErrDriverFailure.Wrap(d.lastErr)
+	}
+
+	deadline, _ := ctx.Deadline()
+	if !deadline.IsZero() {
```
creachadair:
```suggestion
	if deadline, ok := ctx.Deadline(); ok {
```

driver/native/native.go
```diff
@@ -191,6 +226,26 @@ func (d *Driver) Parse(rctx context.Context, src string) (nodes.Node, error) {
 	d.mu.Lock()
 	defer d.mu.Unlock()
 
+	if d.state != stateOK && d.state != stateTimeout {
+		return nil, driver.ErrDriverFailure.Wrap(d.lastErr)
+	}
+
+	deadline, _ := ctx.Deadline()
```
creachadair:
```suggestion
```

driver/native/native_test.go
```diff
@@ -119,3 +133,35 @@ func TestNativeDriverNativeParse_Malformed(t *testing.T) {
 	require.NotNil(err)
 	require.True(derrors.ErrDriverFailure.Is(err))
 }
+
+func TestNativeDriverParse_Timeout(t *testing.T) {
+	require := require.New(t)
+
+	// This test check if the Go driver server can recover from response read timeout.
+	//
+	// For this, we start a mock that will sleep 5 sec before answering any requests.
```
creachadair: We can probably use shorter times—5 sec. is a long time to make a test hang. I'd suggest like 1 sec. and 250ms.

driver/native/native.go
dennwc: I actually thought the same and then stopped myself thinking that encoder might not be JSON. But for now it is, so I'll change it.

driver/native/native_test.go
dennwc: CI is sometimes extremely slow. This may lead to flaky behavior. But we can try.

driver/native/native.go
dennwc: Is a zero deadline useful? If not, what is the point to check it's existence? Is it to account for weird 3rd-party implementations of the `Context`?

driver/native/native_test.go
creachadair: I noticed Travis is very slow to start up, but once it's running it doesn't seem too bad. But if you think it'll cause problems we don't need to change it, it just seems like a long hang.

driver/native/native.go
creachadair: No, it's just cosmetic: It gets rid of an explicitly-ignored return value, which is a red flag to the reader (harmless here, but still made me look). And, it narrows the scope of `deadline` which isn't needed outside this block.

juanjux:
I want to read this later more slowly but LGTM, good hunting!

PR #354, opened 2019-01-31
More robust comment string parsing

juanjux:
- Checks values returned by Index functions for -1 (had a case where it was -1, check failing CI in until this is merged in bblfsh/python-driver/pull/176).
- Trims the text before checking if the prefix and suffix exist with `HasPrefix` and `HasSuffix` to avoid cases where the language can give a wrong position for the start of the comment (or only the line) like happens with Python with some comments where it gives all the whitespace before if they're on their own line and indented.

### DIFF
--- a/uast/transformer/semantic.go
+++ b/uast/transformer/semantic.go
@@ -101,6 +101,19 @@ func CommentText(tokens [2]string, vr string) Op {
 		prefVar:    vr + "_pref",
 		suffVar:    vr + "_suff",
 		indentVar:  vr + "_tab",
+		doTrim:     false,
+	}
+}
+
+func CommentTextTrimmed(tokens [2]string, vr string) Op {
+	return &commentUAST{
+		startToken: tokens[0],
+		endToken:   tokens[1],
+		textVar:    vr + "_text",
+		prefVar:    vr + "_pref",
+		suffVar:    vr + "_suff",
+		indentVar:  vr + "_tab",
+		doTrim:     true,
 	}
 }
 
@@ -127,6 +140,7 @@ type commentElems struct {
 	Prefix     string
 	Suffix     string
 	Indent     string
+	DoTrim     bool
 }
 
 func (c *commentElems) isTab(r rune) bool {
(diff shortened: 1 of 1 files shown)


uast/transformer/semantic.go
```diff
@@ -147,25 +147,36 @@ func (c *commentElems) isTab(r rune) bool {
 }
 
 func (c *commentElems) Split(text string) bool {
-	if !strings.HasPrefix(text, c.StartToken) || !strings.HasSuffix(text, c.EndToken) {
+	trimmed := strings.TrimSpace(text)
```
dennwc: This shouldn't happen automatically. It belongs to a separate op, or should be specified explicitly as a flag.

uast/transformer/semantic.go
```diff
@@ -147,25 +147,36 @@ func (c *commentElems) isTab(r rune) bool {
...
 
-	// find prefix
+	// find prefix (if not already trimmed)
 	i := strings.IndexFunc(text, func(r rune) bool {
 		return !c.isTab(r)
 	})
-	c.Prefix = text[:i]
-	text = text[i:]
+
+	if i == -1 {
```
dennwc: This might be nicer:
```golang
c.Prefix = ""
if i >= 0 {
  c.Prefix = text[:i]
  // ...
}
```

juanjux: Yeah, but is assigns twice in the most common case and I don't trust Golang optimizer much, anyway I also find it better and I don't think it'll hurt much.

uast/transformer/semantic.go
juanjux: Sorry, I rebased and this got fakely outdated. You mean like for example adding a new "trimText" field to `commentUAST` and a new op like `CommentTextTrimmed`?

dennwc: Yes, something like this will be nice :)

But if positions are off the node will fail the token verification for sure. So fixing positions should fix both issues (validation and trim).

juanjux: Done. In the case of Python it works because it only happens with comments on their own line that are (sometimes) indented. In the native AST those doesn't have position, unfortunately, trough in this case it should be probably easy to fix.

uast/transformer/semantic.go
juanjux: Done.

uast/transformer/semantic.go
```diff
@@ -147,25 +161,40 @@ func (c *commentElems) isTab(r rune) bool {
 }
 
 func (c *commentElems) Split(text string) bool {
+	if c.DoTrim {
+		text = strings.TrimLeftFunc(text, func(r rune) bool {
```
dennwc:
```suggestion
		text = strings.TrimLeftFunc(text, unicode.IsSpace)
```

uast/transformer/semantic.go
```diff
@@ -101,6 +101,19 @@ func CommentText(tokens [2]string, vr string) Op {
 		prefVar:    vr + "_pref",
 		suffVar:    vr + "_suff",
 		indentVar:  vr + "_tab",
+		doTrim:     false,
+	}
+}
+
+func CommentTextTrimmed(tokens [2]string, vr string) Op {
+	return &commentUAST{
```
dennwc: You can probably  do something like
```golang
op := CommentTextTrimmed(tokens, vr).(*commentUAST)
op.doTrim = true
```
or add an unexported `newCommentText` with the third argument.

PR #356, opened 2019-02-08
Fix field usage validation when joining objects

dennwc:
Fix a DSL validation bug discovered while working on C# driver. Also added more tests for similar cases.

As mentioned previously, the bug affected the DSL validation and may be a cause of dropped fields, since it gives no error for transforms that forget to specify a specific field that exists in the specific AST node type.

This may break existing drivers that exploited the bug, bypassing the validation by mistake.

### DIFF
--- a/uast/transformer/errors.go
+++ b/uast/transformer/errors.go
@@ -5,6 +5,9 @@ import (
 	"fmt"
 
 	"gopkg.in/src-d/go-errors.v1"
+
+	"gopkg.in/bblfsh/sdk.v2/uast"
+	"gopkg.in/bblfsh/sdk.v2/uast/nodes"
 )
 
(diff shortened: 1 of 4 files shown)


dennwc:
@bzz It's a good question!

Since we are focused on C# right now, I will update the SDK for this driver, and some annotations will stop working since they exploited this bug. I will need some time to fix them.

After this, I plan to go through each driver, update the SDK and `bblfsh-sdk test` it. Since the bug was in the validation code, some annotation might start failing and will need to be fixed as well.

dennwc:
Update: PR needs some cosmetic changes to help in fixing those hidden issues.

dennwc:
OK, should be better now.

PR #360, opened 2019-02-13
Update versions of SDK and bblfshd in the driver skeleton

dennwc:
* Update SDK version for new drivers.
* Update Go version for new drivers and CI
* Update bblfshd version for CI

### DIFF
--- a/etc/skeleton/.travis.yml
+++ b/etc/skeleton/.travis.yml
@@ -1,13 +1,13 @@
 language: go
 
 go:
-  - '1.10'
+  - '1.11'
 
 services:
   - docker
 
 env:
-  - BBLFSHD_VERSION=v2.9.1
+  - BBLFSHD_VERSION=v2.11.7
 
 install:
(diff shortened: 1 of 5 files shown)


etc/skeleton/.travis.yml
```diff
@@ -1,13 +1,13 @@
 language: go
 
 go:
-  - '1.10'
+  - '1.11'
```
creachadair: Should we also continue to test 1.10? I don't know how long we promise compatibility.

bzz: Officially https://github.com/src-d/guide/blob/master/engineering/conventions/go.md#supported-go-versions

But I would suggest deciding on the consistent strategy to get there: right now it's `1.10` basically _everywhere_ in bblfsh. 

So for the purpose of trackable update if I may, I would suggest having an ☂️ issue, listing all the places we want to update the version/make sure it complies with policy and doing that in a single run?

That is, in case we decide that it's a priority ofcourse

dennwc: This is for the driver only, not SDK. So I think it may be okay to support a single version. Most users will see the docker image anyway.

PR #362, opened 2019-02-15
Fix SDK test to run on non-linux os

bzz:
This change consists of
 - documentation for `bblfhs-sdk test`
 - change default test timeout to 1 min
 - always compile `fixtures.test` binary for linux OS, so it could always be run inside a container

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,8 @@ at `$GOPATH/bin/`.
 
 ### Contribute
 
-The SDK provides many templates for language drivers.
-These templates are converted to Go code that ends up in both `bblfsh-sdk` and `bblfsh-sdk-tools` automatically. Use `make` convert templates to code:
+The SDK provides scaffolding templates for creating a new language driver.
+These templates are converted to Go code that ends up in `bblfsh-sdk` tool. Use `make` to update these templates:
 
 ```bash
 $ make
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -150,6 +150,26 @@ For further details of how to construct a language driver,
...
+
+This will:
+ - compile a "test binary" that parses content of the `./fixtures` directory of the driver
+ - create a docker image with all dependencies, native driver and a test binary
+ - run this test binary inside a Docker container, using that image
+
+ Overall, SDK supports 3 different kind of tests for a driver:
+  - UnitTests, parsing content of `./fixtures` and applying UAST transformations. Described above.
+  - Integration tests, using content of `./fixtures/_integration*`
+  - Benchmarks, using content of `./fixtures/bench_*`
```
juanjux: There are also unified tests, the ones starting with `u2_` which implement the same things with a unified struct. Used to test semantic objects.

cmd/bblfsh-sdk/cmd/build.go
```diff
@@ -32,6 +32,8 @@ func (c *BuildCommand) Execute(args []string) error {
 
 const TestCommandDescription = "tests the driver using fixtures"
 
+// TestCommand will compare gold annotations in driver ./fixtures to the
```
juanjux: s/gold/old

driver/fixtures/fixtures.go
```diff
@@ -27,7 +27,7 @@ const Dir = "fixtures"
 const (
 	syntaxErrTestName = "_syntax_error"
 	maxParseErrors    = 3
-	parseTimeout      = time.Minute / 30
+	parseTimeout      = 5 * time.Minute
```
juanjux: Maybe five minutes is too much.

dennwc: Agree, 1 minute should be enough.

README.md
bzz: So far I only documented the things I know and understand :) 

If you could provide an exact wording here - I'll be happy to add! 

Otherwise will keep it as is and update it later, when I stumble up on `u2_` and understand their purpose better.

cmd/bblfsh-sdk/cmd/build.go
bzz: Hm, I think usage of the word "gold" is common when referring to a "universal source of truth".
Was meant in that sense only. Otherwise need to clarify what are old and what are new.

If you have an exact wording - please post here (btw I find using a [new github suggestions feature](https://twitter.com/github/status/1053360435074949120?lang=en) to be faster for such cases)

driver/fixtures/fixtures.go
bzz: Done in b1de985

cmd/bblfsh-sdk/cmd/build.go
bzz: Tried to make it a bit more clear in b1de985

bzz:
Thank you for prompt feedback! All adressed in b1de985 and is ready for another round @dennwc @juanjux

PR #363, opened 2019-02-16
Add few node helpers

dennwc:
Add a `ChildrenCount` helper (requested by https://github.com/bblfsh/client-go/issues/109) and relax argument constraints for `PositionsOf`, `RolesOf` and `TokenOf` (see https://github.com/bblfsh/client-go/issues/100).

Fixes https://github.com/bblfsh/client-go/issues/109.

### DIFF
--- a/uast/nodes/node.go
+++ b/uast/nodes/node.go
@@ -948,3 +948,29 @@ func UniqueKey(n Node) Comparable {
 		return unkPtr(ptr)
 	}
 }
+
+// ChildrenCount reports the number of immediate children of n. If n is an Array this is
+// the length of the array. If n is an Object, each object in a field of n counts as
+// one child and each array is counted as its length.
+func ChildrenCount(n Node) int {
+	switch n := n.(type) {
+	case nil:
+		return 0
+	case Value:
+		return 0
+	case Array:
+		return len(n)
+	case Object:
+		c := 0
+		for _, v := range n {
+			switch v := v.(type) {
+			case Object:
+				c++
+			case Array:
+				c += len(v)
+			}
+		}
+		return c
+	}
+	return 0
+}
--- a/uast/uast.go
+++ b/uast/uast.go
@@ -182,7 +182,13 @@ func AsPosition(m nodes.Object) *Position {
 }
 
 // PositionsOf returns a complete positions map for the given UAST node.
-func PositionsOf(m nodes.Object) Positions {
+// The function will return nil for non-object nodes like arrays and values. To get
+// positions for these nodes, PositionsOf should be called on their parent node.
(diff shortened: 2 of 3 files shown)


uast/nodes/node.go
```diff
@@ -948,3 +948,28 @@ func UniqueKey(n Node) Comparable {
 		return unkPtr(ptr)
 	}
 }
+
+// ChildrenCount returns a number of immediate children nodes. For arrays it will be the
+// length of an array. For objects, all child Object and Array nodes in fields will be counted.
```
creachadair: I was confused (until I read the code) about whether this means the Array is counted as 1 (for itself) or for its children. May I suggest

```
// ChildrenCount reports the number of immediate children of n. If n is an Array this is
// the length of the array. If n is an object, each object or array in a field of n counts as 
// one child.
```

or words to that effect?

uast/uast.go
```diff
@@ -182,7 +182,11 @@ func AsPosition(m nodes.Object) *Position {
 }
 
 // PositionsOf returns a complete positions map for the given UAST node.
-func PositionsOf(m nodes.Object) Positions {
+func PositionsOf(n nodes.Node) Positions {
```
creachadair: Please update the comment to document that this is empty for a non-Object. Previously it was clear from the type signature, but no longer—and intuitively it's not obvious that an array wouldn't also have positions (e.g., the start and end of its span). Not saying it should, but if I asked for the position of an arbitrary node I would be surprised that it should't work for an Array.

uast/uast.go
```diff
@@ -216,7 +220,11 @@ func RoleList(roles ...role.Role) nodes.Array {
 }
 
 // RolesOf is a helper for getting node UAST roles (see KeyRoles).
-func RolesOf(m nodes.Object) role.Roles {
+func RolesOf(n nodes.Node) role.Roles {
+	m, ok := n.(nodes.Object)
```
creachadair: As 185. Is there some reason for this relaxation more generally?

bzz:
CI seems to be failing for some reason

uast/nodes/node.go
dennwc: Changed the comment, PTAL.

uast/uast.go
dennwc: Done.

uast/uast.go
dennwc: The only reason to relax this requirement is that all XPath query and any manual node manipulations will usually return a `nodes.Node`, and the user have to type assert it each time to use those functions.

dennwc:
CI passes, fixed comments and added test cases. Ready for another round.

dennwc:
@bzz PTAL

PR #369, opened 2019-02-21
Add Drop operation to Fields

dennwc:
As requested in https://github.com/bblfsh/cpp-driver/pull/26, add an option to drop a field. It will work the same way as `Optional: "x", Op: Any()`, except that it won't create a variable for `Optional`.

### DIFF
--- a/uast/transformer/ops.go
+++ b/uast/transformer/ops.go
@@ -746,11 +746,25 @@ func (op opScope) Construct(st *State, n nodes.Node) (nodes.Node, error) {
 // Field is an operation on a specific field of an object.
 type Field struct {
 	Name string // name of the field
-	// Optional can be set to make a field optional. Provided string is used as a variable name to the state of the field.
-	// Note that "optional" means that the field may not exists in the object, and it does not mean that the field can be nil.
+	// Optional can be set to make a field optional. Provided string is used as a variable
+	// name to the state of the field. Note that "optional" means that the field may not
+	// exists in the object, and it does not mean that the field can be nil.
 	// To handle nil fields, see Opt operation.
 	Optional string
-	Op       Op // operation used to check/construct the field value
+	// Drop the field if it exists. Op should be set and it will be called to check the
+	// value before dropping it. If the check fails, the whole transform will be canceled.
+	//
+	// Please note that you should avoid dropping fields with Any unless there is no
+	// reasonable alternative.
+	Drop bool
+	Op   Op // operation used to check/construct the field value
+}
+
+// Desc returns a field descriptor.
+func (f Field) Desc() FieldDesc {
+	d := FieldDesc{Optional: f.Optional != "" || f.Drop}
+	d.SetValue(f.Op)
+	return d
 }
 
 var _ ObjectOp = Fields{}
@@ -770,9 +784,7 @@ func (Fields) Kinds() nodes.Kind {
 func (o Fields) Fields() (FieldDescs, bool) {
 	fields := make(FieldDescs, len(o))
 	for _, f := range o {
-		fld := FieldDesc{Optional: f.Optional != ""}
-		fld.SetValue(f.Op)
-		fields[f.Name] = fld
+		fields[f.Name] = f.Desc()
 	}
 	return fields, true
 }
@@ -800,11 +812,11 @@ func (o Fields) CheckObj(st *State, n nodes.Object) (bool, error) {
 			if err := st.SetVar(f.Optional, nodes.Bool(ok)); err != nil {
 				return false, errKey.Wrap(err, f.Name)
 			}
-			if !ok {
-				continue
-			}
 		}
 		if !ok {
+			if f.Optional != "" || f.Drop {
+				continue
+			}
 			if errorOnFilterCheck {
 				return filtered("field %+v is missing in %+v\n%+v", f, n, o)
 			}
--- a/uast/transformer/ops.go
+++ b/uast/transformer/ops.go
@@ -751,8 +751,9 @@ type Field struct {
 	// exists in the object, and it does not mean that the field can be nil.
 	// To handle nil fields, see Opt operation.
 	Optional string
-	// Drop the field if it exists. Op should be set and it will be called to check the
-	// value before dropping it. If the check fails, the whole transform will be canceled.
+	// Drop the field if it exists. Optional is implied, but the variable won't be created
+	// in this case. Op should be set and it will be called to check the value before
+	// dropping it. If the check fails, the whole transform will be canceled.
 	//
 	// Please note that you should avoid dropping fields with Any unless there is no
 	// reasonable alternative.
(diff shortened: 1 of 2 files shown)


uast/transformer/ops.go
```diff
@@ -746,11 +746,23 @@ func (op opScope) Construct(st *State, n nodes.Node) (nodes.Node, error) {
...
-	// Note that "optional" means that the field may not exists in the object, and it does not mean that the field can be nil.
+	// Optional can be set to make a field optional. Provided string is used as a variable
+	// name to the state of the field. Note that "optional" means that the field may not
+	// exists in the object, and it does not mean that the field can be nil.
 	// To handle nil fields, see Opt operation.
 	Optional string
-	Op       Op // operation used to check/construct the field value
+	// Drop the field if it exists. If Op is set, it will be called to check the value
+	// before dropping it. If the check fails, the whole transform will be canceled.
+	// This is the preferred way for dropping fields. Avoid using Drop without Op.
```
creachadair: Would it make sense to consider this an error in the validation spec? If we make it always an error to have drop=true with op=nil, that would make tests fail consistently if we forget to set them.

Not crucial, IMO, but I worry a little about how this would manifest in tests.

juanjux: I don't really understand the "Avoid using Drop without Op" part. Does it means that `Op: Any()` should be used anyway? If that's the case: could it be implicit?

dennwc: It sets `Op: Any()` by default if it's not specified and `Drop` is set.

The comment is there because we should avoid `Any`... at any cost :) So it's better to set `Drop: true` and `Op: Bool(true)` for example, so you explicitly specify when the field can be dropped.

dennwc: No, it's not an error, it will just assume `Op: Any()` in this case.

See https://github.com/bblfsh/sdk/pull/369#discussion_r259027330

dennwc:
OK, I see that the comment is confusing. Any ideas on how to make it more clear?

uast/transformer/ops.go
```diff
@@ -853,7 +866,11 @@ func (o Fields) ConstructObj(st *State, obj nodes.Object) (nodes.Object, error)
 				continue
 			}
 		}
-		v, err := f.Op.Construct(st, nil)
+		op := f.Op
+		if f.Drop && op == nil {
```
juanjux: So the comment above should be "avoid using drop with an op", shouldn't it? Maybe `f.Drop && op != nil` should be an error in this case.

uast/transformer/ops.go
juanjux: I think using Drop is specific enough. The comment could add "Please note that you should avoid dropping fields in general unless there is not reasonable alternative".

creachadair: Right, I guess what I'm suggesting is maybe we should require the `Any()` to be added explicitly if that's what you want—then there's no chance we get it "wrong".

uast/transformer/ops.go
dennwc: No, it's actually a preferred way of doing it. It reads like this: "drop if the field not exists, or if it matches the Op".

uast/transformer/ops.go
dennwc: Yeah, this sounds like a better alternative, although a bit more verbose. But it's fine for the "wrong" case of using `Any` to drop things.

creachadair: Yeah, since we suggest not using Drop unless necessary, I think the extra friction will actually be useful in this situation :)

uast/transformer/ops.go
juanjux: I just realised that `Drop` with an op could be used to restrict the kind of values dropped instead of always dropping the field (for example, 'drop this field if the string value is "foo"') so I said nothing :)

uast/transformer/ops.go
juanjux: Now that I understand how the Op could be used, I would rephrase as:

```
// Drop the field if it exists (so Optional is implied). If Op is set, it will be called to check the value
// and only drop the field if the value matches it. If it's not set, the field will always be dropped
// but please note that you should avoid dropping fields in this way if there is no
// reasonable alternative.
```

dennwc:
Ready for another review round.

uast/transformer/ops.go
```diff
@@ -746,11 +746,25 @@ func (op opScope) Construct(st *State, n nodes.Node) (nodes.Node, error) {
...
 	Name string // name of the field
-	// Optional can be set to make a field optional. Provided string is used as a variable name to the state of the field.
-	// Note that "optional" means that the field may not exists in the object, and it does not mean that the field can be nil.
+	// Optional can be set to make a field optional. Provided string is used as a variable
+	// name to the state of the field. Note that "optional" means that the field may not
+	// exists in the object, and it does not mean that the field can be nil.
 	// To handle nil fields, see Opt operation.
 	Optional string
-	Op       Op // operation used to check/construct the field value
+	// Drop the field if it exists. Op should be set and it will be called to check the
```
juanjux: I would add:

+ Optional is implied (even if its implicitly told by "if it exists").

uast/transformer/ops.go
```diff
@@ -746,11 +746,25 @@ func (op opScope) Construct(st *State, n nodes.Node) (nodes.Node, error) {
...
+	// value before dropping it. If the check fails, the whole transform will be canceled.
```
juanjux: Maybe "the whole transform will be cancelled like if you used Any()" which is also didactic.

uast/transformer/ops.go
```diff
@@ -746,11 +746,25 @@ func (op opScope) Construct(st *State, n nodes.Node) (nodes.Node, error) {
...
+	//
+	// Please note that you should avoid dropping fields with Any unless there is no
```
juanjux: "you should avoid dropping fields with Any (explicit or implicit) unless there is no..."

juanjux:
More nitpicking trying to look at it from the perspective or a new, clueless and confused DSL user (because I've been there haha).

uast/transformer/ops.go
dennwc: See the comment thread above: https://github.com/bblfsh/sdk/pull/369#discussion_r259029902

I remove the implicit `Any`.

juanjux:
Perfection has been reached 👌

bzz:
Regarding the discussion in https://github.com/bblfsh/sdk/pull/369#discussion_r259029902 - would you be so kind to help me understand the expected behavior of - `{Name: "specifiers", Drop: true, Op: Arr()},` is that suppose to match&drop only empty array?

Because from what I can tell - right now this results in `check: unused field(s) on node ImportDeclaration: specifiers` 😕

juanjux:
Could it be the case that there is a `specifiers` field that is not an empty array (so the `Drop` wouldn't match) and it's not being handled?

bzz:
It could, but I wanted to verify the expectations here: `{Name: "specifiers", Drop: true, Op: Arr()}` is supposed to match&drop `specifiers` _only_ in case it's an empty array - would that be correct?

juanjux:
I think it is but less wait for the DSL guru @dennwc to confirm this.

dennwc:
Yes, it's correct. If it's not an empty array the whole operation will be reverted.

PR #372, opened 2019-03-12
Optimize DSL

dennwc:
This PR incorporates two optimizations found during recent profiling session.

First optimization changes `nodes.Equal` function to be more efficient. The function itself is pretty fast, but it executes so often that it dominates the execution time of the DSL (node filtering). The change inlines type-specific checks and defines a specialized `NodeEqual` function that omits interface convertions (`runtime.convI2I`).

The second optimization changes the `transform.FieldDescs` (field descriptors) from map to a struct with a slice. This increases the performance of range loops over this structure, resulting in ~30% DSL performance increase. The structure makes a difference, because `FieldDescs` is used during node matching for partial transformations that are used more often in recent driver. DSL cannot optimize the filtering for this specific transform, so it end up testing all nodes, thus map iteration give a considerable performance penalty for this case.

Benchmark comparison for C# driver with this change applied:
```
benchmark                                            old ns/op      new ns/op      delta
BenchmarkCsharpDriver/transform/binary_search-4      153618333      76028455       -50.51%
BenchmarkCsharpDriver/transform/parser_context-4     2385605943     1153439690     -51.65%
```

### DIFF
--- a/uast/nodes/node.go
+++ b/uast/nodes/node.go
@@ -45,20 +45,30 @@ func Equal(n1, n2 External) bool {
 			return n1 == n2
 		}
 	case Object:
-		n2, ok := n2.(Object)
-		if !ok || len(n1) != len(n2) {
+		if n2, ok := n2.(Object); ok {
+			if len(n1) != len(n2) {
+				return false
+			}
+			if pointerOf(n1) == pointerOf(n2) {
+				return true
+			}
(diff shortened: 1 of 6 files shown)


dennwc:
Merging this one, since it's blocking driver update.

@bzz Let me know if you have any comments.

PR #373, opened 2019-03-12
Allow to run driver UAST transform benchmark on the host

dennwc:
Currently, all benchmarks including the ones for transformation pipeline run the driver instance to parse files. This prevents running those benchmarks on the host since driver runtime is necessary to parse files.

Instead, rewrite transformation pipeline benchmarks to use AST files generated by tests (fixture files). Those files will always be there since we use them to test the output of the driver.

This allows running this specific benchmark on the host, making results more stable.

A side effect is that, due to the way how `uast.RolesOf` is implemented, it will always append `Unannotated` role to all nodes that do not define roles and are not part of the Semantic UAST schema.

YAML encoder that is used to write fixtures in tests uses this function to dump native AST. As a result, all AST nodes end up with an `Unannotated` role and a new `@role` field. This breaks the transformation tests because of the unknown field that should not exist for Native parsing mode.

This PR also adds an option for the YAML UAST encoder to disable roles for native AST. Test are changed according, so drivers will need to re-generate native fixtures after the next SDK update.

### DIFF
--- a/driver/fixtures/fixtures.go
+++ b/driver/fixtures/fixtures.go
@@ -14,7 +14,6 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"gopkg.in/bblfsh/sdk.v2/driver"
-	"gopkg.in/bblfsh/sdk.v2/protocol/v1"
 	"gopkg.in/bblfsh/sdk.v2/uast"
 	"gopkg.in/bblfsh/sdk.v2/uast/nodes"
 	"gopkg.in/bblfsh/sdk.v2/uast/transformer/positioner"
@@ -112,12 +111,7 @@ func (s *Suite) RunTests(t *testing.T) {
 }
 
 func (s *Suite) RunBenchmarks(b *testing.B) {
-	b.Run("transform", func(b *testing.B) {
-		s.benchmarkTransform(b, false)
-	})
-	b.Run("transform-legacy", func(b *testing.B) {
-		s.benchmarkTransform(b, true)
-	})
+	b.Run("transform", s.benchmarkTransform)
 	b.Run("fixtures", s.benchmarkFixtures)
 }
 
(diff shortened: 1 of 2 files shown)


bzz:
Sounds like a good idea. 

Could you please help understand if this does change anything on UI level e.g how to run these benchmarks, etc? BTW if that is already documented somewhere - would yo be so kind to point me to that? Thanks!

dennwc:
Currently, you can run benchmarks in Docker with `bblfsh-sdk test -bench`. I'm not aware of any documentation that mentions this, except for the help in CLI.

This PR allows to run it as usual: `go test -run=NONE -bench=. ./driver/fixtures`. But because some of the benchmarks still try to run the driver, this command will run only a few of them and will fail for others.

Until we find a good way to detect if the test is run in Docker or not, you can specify a specific benchmark to run: `go test -run=NONE -bench=BenchmarkXxxxDriver/transform ./driver/fixtures`.

creachadair:
> Until we find a good way to detect if the test is run in Docker or not

In a script, I think it should suffice to check `[[ -f /.dockerenv ]]`. Of course that could be fooled, but I think it's good enough for someone manually running benchmarks.

dennwc:
@creachadair That would be great. Let me check it, and I'll push an update to enable it for other benchmarks.

dennwc:
Works like a charm! The usual `go test -run=NONE -bench=. ./...` is safe now.

Ready for another round.

driver/fixtures/fixtures.go
```diff
@@ -44,6 +44,11 @@ type DockerConfig struct {
 	Image string
 }
 
+func runsInDocker() bool {
```
creachadair: What you wrote here is completely fine, but I would consider making it specific:

```go
func skipInDocker(t *testing.T) {
  t.Helper()
  if _, err := os.Stat("/.dockerenv"); err == nil {
    t.SkipNow()
  }
}
…
skipInDocker(t)
```

It's really up to you, but it's nice to reduce the conditional structure inside the Test functions.

creachadair: Oh. Now I see you do this for both *testing.T and *testing.B, so nevermind.

driver/fixtures/fixtures.go
```diff
@@ -410,6 +423,9 @@ func isBench(name, ext string) (string, bool) {
 
 func (s *Suite) benchmarkFixtures(b *testing.B) {
 	b.StopTimer()
+	if !runsInDocker() {
```
creachadair: Maybe do this first? (Not that it really matters when you stop the timer, but there's no point in doing anything for a benchmark that isn't going to run)

dennwc:
Merging this one, since it's blocking driver update.

@bzz Let me know if you have any comments.

PR #374, opened 2019-03-13
build: propagate tag inside build container

bzz:
This change allows make build system inside the container to get a version number that will be embedded in go binary and override the default `dev-<shar>`

Fixes https://github.com/bblfsh/javascript-driver/issues/55

TODO:
 - [x] update bindata


bzz:
CI is failing with

```
	discovery_test.go:28: 
			Error Trace:	discovery_test.go:28
			Error:      	Received unexpected error:
			            	GET https://api.github.com/orgs/bblfsh/repos?page=1&per_page=100&type=public: 403 API rate limit exceeded for 35.184.226.236. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) [rate reset in 25m28s]
```

that is not relevant to the change. Bindate update was force-pushed

PR #375, opened 2019-03-14
Fix edge case for comment parsing

dennwc:
Comment parser does a correct check for the case when only stylistic characters are present, but it does not set the prefix correctly for this case.


bzz:
If this would result in any fixture changes on the driver update, I suggest mentioning this in SDK release nodes to communicate the expectations to driver developers/maintainers as clearly as possible.

dennwc:
Yes, it will, but only for this specific edge case. Will update all the drivers and make sure to add a note in the release notes.

PR #378, opened 2019-03-15
Use Go scripts instead of bblfsh-sdk binary

### LINKED ISSUE #325: bblfsh-sdk does not account for the driver's SDK version
Currently, if the version of the driver and `bblfsh-sdk` tool doesn't match, the tool will always try to override "changed" files. This change will break the driver's CI since it will install an older SDK version, as required by the manifest.

Instead, the tool should get the requested SDK version in the manifest (we have this functionality already) and compare it with it's own version. The tool should fail with a meaningful message if major+minor versions are not the same.

An alternative is to automatically execute `go run ./vendor/gopkg.in/bblfsh/sdk.v2/cmd/bblfsh-sdk/main.go` when running the tool if it exists in the `vendor` directory.

CC @juanjux

dennwc:
There are quite a few problems when using `bblfsh-sdk` tool to build/test drivers.

The most significant one is that `bblfsh-sdk` binary may be built with a different SDK version than the driver itself. It will cause the tool to update driver files, while the driver will fail to build using those
files either locally or in CI, where the correct tool version will be available.

Since we have a copy of SDK in the driver's `vendor` directory anyway, it's better to use it to build the driver. This is done in this PR by a few small Go scripts in the driver directory. They will call the build or test command the same way as the `bblfsh-sdk` tool does.

The change does not affect the `bblfsh-sdk` tool itself - it can still be used to initialize new drivers or update existing ones.

Resolves #325


bzz:
A quick question on

> This is done in this PR by a few small Go scripts in the driver directory. They will call the build or test command the same way as the bblfsh-sdk tool does.

could you please help to understand what is the rationale/pros for having new scripts in root instead of just `go run vendor/gopkg.in/bblfsh/sdk.v2/cmd/bblfsh-sdk/main.go update` ? Thanks!

juanjux:
Note that the docs should also be updated, there is a references to bblfsh-sdk udpate in:

https://doc.bblf.sh/writing-a-driver/babelfish-sdk.html

dennwc:
@bzz CI is not failing, it's code coverage part because the update functionality was moved into a separate package.

Regarding `vendor` - you can still use that command, but it's too long for my taste :) `go run update.go` is much simpler and easier to remember for newcomers. Both commands do exactly the same though.

build/sdk_update.go
```diff
@@ -0,0 +1,316 @@
...
+	c.changes++
+	c.opt.Warningf("managed file changed %q\n", file)
+}
+
+func escapeShield(text interface{}) string {
+	return strings.Replace(fmt.Sprintf("%s", text), "-", "--", -1)
+}
+
+func expName(s string) string {
+	if len(s) == 0 {
```
creachadair: Optional suggestion:

```go
r, n := utf8.DecodeRuneInString(s)
if n == 0 {
  return s
}
return string(unicode.ToUpper(r))+s[n:]
```

build/sdk_update.go
```diff
@@ -0,0 +1,316 @@
...
+
+	c.changes++
+	c.opt.Warningf("managed file changed %q\n", file)
+}
+
+func escapeShield(text interface{}) string {
+	return strings.Replace(fmt.Sprintf("%s", text), "-", "--", -1)
+}
+
+func expName(s string) string {
```
creachadair: What does the name of this function mean? It looks like it capitalizes the first rune in the string, but the name doesn't suggest that.

build/sdk_update.go
```diff
@@ -0,0 +1,316 @@
...
+func (c *updater) processFileAsset(name string, overwrite bool) error {
+	content := skeleton.MustAsset(name)
+	info, _ := skeleton.AssetInfo(name)
+
+	name = fixGitFolder(name)
+	return c.writeFile(filepath.Join(c.root, name), content, info.Mode(), overwrite)
+}
+
+var funcs = map[string]interface{}{
+	"escape_shield": escapeShield,
```
creachadair: Could this be named just `"escape"` or maybe `"quote"`? 

(Is this name something pre-existing?)

build/sdk_update.go
```diff
@@ -0,0 +1,316 @@
...
+
+	name = fixGitFolder(name)
+	file := filepath.Join(c.root, name[:len(name)-len(tplExt)])
+
+	buf := bytes.NewBuffer(nil)
+	if err := t.Execute(buf, v); err != nil {
+		return err
+	}
+
+	info, _ := skeleton.AssetInfo(name)
```
creachadair: We should probably not ignore these errors, but it'd be fine to panic for them if indeed this is the mirror of `skeleton.MustAsset` for the file info instead of the content:

```go
func mustAssetInfo(name string) os.FileInfo {
  fi, err := skeleton.AssetInfo(name)
  if err != nil {
    log.Panicf("Missing asset info for %q: %v", name, err)
  }
  return fi
}
```

or whatever.

(In fact we might think about adding that to the API for the `skeleton` package separately, but for now I think it'd be fine to host it here)

build/sdk_update.go
dennwc: Yes, actually the whole file was just moved :)

I'm not sure if something will break in the templating engine if I change those names.

build/sdk_update.go
```diff
@@ -0,0 +1,316 @@
...
+		if out, err := git.CombinedOutput(); err != nil {
+			c.opt.Warningf("cannot add a file to git: %v\n%s", err, string(out))
+		}
+	}
+	return nil
+}
+
+func (c *updater) notifyMissingFile(file string) {
+	if !c.opt.DryRun {
+		c.opt.Noticef("creating file %q\n", file)
```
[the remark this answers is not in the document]
dennwc: It will actually notify in the dry run and warn in the normal mode. So I think it's doing exactly what it should.

build/sdk_update.go
```diff
@@ -0,0 +1,316 @@
...
+
+	if bytes.Compare(original, content) == 0 {
+		return nil
+	}
+
+	c.notifyChangedFile(file)
+	return c.doWriteFile(file, content, m)
+}
+
+func (c *updater) doWriteFile(file string, content []byte, m os.FileMode) error {
```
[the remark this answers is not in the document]
dennwc: Makes sense, changed.

build/sdk_update.go
```diff
@@ -0,0 +1,316 @@
...
+	if _, err = os.Stat(file); err == nil {
+		return errors.New("trying to overwrite the manifest")
+	}
+	info, _ := skeleton.AssetInfo(name)
+
+	f, err := os.OpenFile(file, os.O_RDWR|os.O_CREATE|os.O_TRUNC, info.Mode())
+	if err != nil {
+		return err
+	}
+	defer f.Close()
```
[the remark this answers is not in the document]
dennwc: It is safe for files. In other cases, I will have to copy-paste the `f.Close` to each error return by hand.

build/sdk_update.go
```diff
@@ -0,0 +1,316 @@
...
+
+	Debugf   PrintfFunc
+	Noticef  PrintfFunc
+	Warningf PrintfFunc
+}
+
+// ErrChangesRequired is returned by SDKUpdate in DryRun mode when changes are required.
+var ErrChangesRequired = errors.New("changes are required")
+
+// GenerateManifest writes the manifest file to a root driver directory.
```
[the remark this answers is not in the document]
dennwc: I think this is unnecessary. If the root is relative, all paths will be relative, obviously.

build/sdk_update.go
```diff
@@ -0,0 +1,316 @@
...
+
+type updater struct {
+	root    string
+	opt     UpdateOptions
+	context map[string]interface{}
+	changes int
+}
+
+// PrintfFunc is a logging function type similar to log.Printf.
+type PrintfFunc func(format string, args ...interface{}) (int, error)
```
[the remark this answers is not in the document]
dennwc: Good idea!

dennwc:
Ready for another round.

Added one more change to allow running UAST transformation tests on the host as well and dropped some unused files.

build/sdk_update.go
```diff
@@ -0,0 +1,320 @@
...
+)
+
+func genEnvBool(key string) bool {
+	str := os.Getenv(key)
+	if str == "" {
+		return false
+	}
+	v, err := strconv.ParseBool(str)
+	if err != nil {
+		panic(err)
```
creachadair: Hmm. I thought about proposing this solution, and didn't on the grounds that optional settings in the environment shouldn't be able to crash the program. I think it's defensible, but would suggest a more detailed error message:

```go
log.Panicf("Invalid value %q for environment flag %q: %v", v, key, err)
```

or words to that effect. Because, importantly, if this _does_ happen, someone is going to have to figure out what to fix in their config files.

build/sdk_update.go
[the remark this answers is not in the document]
creachadair: It happens to work for an `*os.File`, but isn't guaranteed by the interface _or_ promised by the implementation. Since there are only two cases here, what I would recommend is:

```go
err := t.Execute(f, map[string]string{"Language": language})
cerr := f.Close()
if err != nil {
   return err
}
return cerr
```

That doesn't always work nicely if there are multiple routes, but in this case the terminus is close. More generally what I would advise in cases where there isn't an easy resolution is:

```go
func F(name string) (err error) {
    f, err := os.OpenFile(…)
    …
    defer func() {
       if cerr := f.Close(); cerr != nil && err == nil {
         err = cerr
       }
    }()
}
```

which is admittedly not pretty, but avoids the unsupported requirement. (Here, clearly, we don't need that).

bzz:
> Added one more change to allow running UAST transformation tests on the host as well and dropped some unused files.

this sounds really cool for a driver development workflow! 🚀 
@dennwc shall we document how to run tests on host either here, or at least in SDK's README? 

Asking as may be I have missed something, but by only looking at a516dfbbc6d48a996592f93d9d0fc275fb117a69 it does not strike me obvious how to do that yet.

build/sdk_update.go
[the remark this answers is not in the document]
dennwc: Sure, it isn't exactly mentioned in interface docs, but if you check, almost every type actually can be closed twice, not only `os.File`. Same is true for Go libraries in the wild. It's a very common pattern.

Anyway, copy-pasted it for each return. Expect more FD leak bugs in the future :confused:

dennwc:
Updated README and addressed remaining feedback. Ready for another round.

build/sdk_update.go
```diff
@@ -0,0 +1,322 @@
...
+
+	if bytes.Compare(original, content) == 0 {
+		return nil
+	}
+
+	c.notifyChangedFile(file)
+	return c.writeFile(file, content, m)
+}
+
+func (c *updater) writeFile(file string, content []byte, m os.FileMode) error {
```
creachadair: Actually, I realized there's another potential simplification option, which might also address your concerns about control flow:

```go
func (c *updater) writeFile(file string, content []byte, m os.FileMode) error {
	if c.opt.DryRun {
		return nil
	} else if err := os.MkdirAll(filepath.Dir(file), 0755); err != nil {
		return err
	}
	err := ioutil.WriteFile(file, content, m)
	if err != nil {
		return err
	}
	c.opt.Debug.printf("File %q has been written\n", file)
	rel, err := filepath.Rel(c.root, file)
	if err != nil {
		return err
	}
	if !strings.HasPrefix(rel, ".git"+string(filepath.Separator)) {
		git := exec.Command("git", "add", rel)
		git.Dir = c.root
		if out, err := git.CombinedOutput(); err != nil {
			c.opt.Warning.printf("cannot add a file to git: %v\n%s", err, string(out))
		}
	}
	return nil
}
```

What do you think?

dennwc: This looks better from both points of view :)

Issue #380, opened 2019-03-16
driver failure: native driver is not running


smacker:
Try to parse file from this issue: https://github.com/bblfsh/python-driver/issues/182

The driver is broken after it and doesn't get restarted. For example, it is broken right now on https://play.bblf.sh/ and parsing of python files doesn't work anymore.

bzz:
@dennwc as soon as https://github.com/bblfsh/bblfshd/releases/tag/v2.12.0 is released, here are results using it

```
$ docker run -it --name bblfshd --privileged -p 9432:9432 bblfsh/bblfshd:v2.12.0-drivers --log-level=debug

$ wget https://gist.githubusercontent.com/smacker/7a618e37997b7ed9d3b688bea04561f1/raw/3448e22a1b884f82fc513f9cc3d2334af4544dba/decoder.py
$ bblfsh-cli decoder.py
driver failure: invalid character 'I' looking for beginning of value

$ docker exec bblfshd bblfshctl status
+----------+----------------+---------------+---------+--------+
| LANGUAGE | SUCCESS/FAILED | STATE/DESIRED | WAITING | EXITED |
+----------+----------------+---------------+---------+--------+
| python   | 0/1            | 1/1           | 0       | 0      |
+----------+----------------+---------------+---------+--------+
Response time 10.8µs

$ bblfsh-cli ./python-driver/fixtures/while.py
driver failure: native driver is not running

$ docker exec bblfshd bblfshctl status
+----------+----------------+---------------+---------+--------+
| LANGUAGE | SUCCESS/FAILED | STATE/DESIRED | WAITING | EXITED |
+----------+----------------+---------------+---------+--------+
| python   | 0/2            | 1/1           | 0       | 0      |
+----------+----------------+---------------+---------+--------+
Response time 13.7µs
```

TL;DR: latest python driver still hits https://github.com/bblfsh/python-driver/issues/182 and this issue is still reproducible.

Server logs
```
[2019-03-26T23:34:32Z]  INFO server listening in /tmp/rpc.sock (unix) id=01d6y5ny1g0v3hepkbkypcc9gj language=python
[2019-03-26T23:34:33Z]  INFO new driver instance started bblfsh/python-driver:latest (01d6y5ny1g0v3hepkbkypcc9gj)
[...]

PR #381, opened 2019-03-19
Restart native driver in case it crashes

dennwc:
It turned out we don't have any logic to recover from a native driver crash (cool, right?).

`bblfshd` is of no help here since the Go driver server (the process that `bblfshd` starts in the container) usually survives the crash. `bblfshd` still thinks that the request failure was temporary and sends new requests, while driver server is not smart enough to restart the native driver process.

This change allows Go driver server to restart the native driver after a crash. As a bonus, it fixes an FD leak on driver close.

Logic is the following: if the driver server sees the end of `stdout` when waiting for a response, it assumes that the driver crashed, restarts it and returns an error to the client. The request of a next client (with a different file) may succeed this way.

Addresses #380 (need to check)

### DIFF
--- a/driver/native/native.go
+++ b/driver/native/native.go
@@ -286,8 +286,9 @@ func (d *Driver) Parse(rctx context.Context, src string) (nodes.Node, error) {
 
 	r, err := d.readResponse(ctx)
 	if err == io.EOF {
-		// driver just died; try to restart it once
+		// driver just died; we don't care about exit code
 		<-d.cmdErr
+		// try to restart it once
 		if err := d.Start(); err != nil {
 			return nil, driver.ErrDriverFailure.Wrap(err, "driver restart failed")
 		}
@@ -329,11 +330,8 @@ func (d *Driver) Close() error {
 	}
 	timeout := time.NewTimer(closeTimeout)
 	select {
-	case err := <-d.cmdErr:
+	case <-d.cmdErr: // don't care about exit code
 		timeout.Stop()
-		if err != nil {
-			last = err
-		}
 	case <-timeout.C:
 		d.cmd.Process.Kill()
 	}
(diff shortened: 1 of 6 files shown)


driver/native/native.go
```diff
@@ -249,20 +281,19 @@ func (d *Driver) Parse(rctx context.Context, src string) (nodes.Node, error) {
...
-		return nil, driver.ErrDriverFailure.Wrap(fmt.Errorf("error: %v; %s", err, string(raw)))
+		return nil, err
 	}
 
 	r, err := d.readResponse(ctx)
+	if err == io.EOF {
+		// driver just died; try to restart it once
+		<-d.cmdErr
+		if err := d.Start(); err != nil {
+			return nil, driver.ErrDriverFailure.Wrap(err, "driver restart failed")
```
juanjux: Maybe also add the driver name (I'm not sure if some previous log related to this failure would print it).

driver/native/native.go
```diff
@@ -249,20 +281,19 @@ func (d *Driver) Parse(rctx context.Context, src string) (nodes.Node, error) {
...
-			// stream is broken on both sides, cannot get additional info
-			return nil, driver.ErrDriverFailure.Wrap(err)
-		}
-		return nil, driver.ErrDriverFailure.Wrap(fmt.Errorf("error: %v; %s", err, string(raw)))
+		return nil, err
 	}
 
 	r, err := d.readResponse(ctx)
+	if err == io.EOF {
+		// driver just died; try to restart it once
```
juanjux: Could be interesting, maybe for future versions, to wait a little before restarting it and increase that wait on N restart tries until failing completely.

driver/native/native.go
dennwc: `bblfshd` should take care of this (it adds the language to the log).

driver/native/native.go
dennwc: Good idea! Added a note. `bblfshd` already has this logic, but it may be useful here as well.

Issue #384, opened 2019-03-21
Grpc: received message larger than max (5797446 vs. 4194304)


bzz:
Discovered as part of the https://github.com/bblfsh/bblfshd/issues/268#issuecomment-475046348

Some drivers (js, go, java, ruby) fail to parse big files with ``

This should have been fixed in https://github.com/bblfsh/sdk/pull/329 but for some reason seems to happen again.

From the logs it's not possible to identify the affected driver actual version numbers, until we have https://github.com/bblfsh/bblfshd/issues/269

bzz:
Test on 5.8 mb file - it works now 🎉 

```bash
docker run --rm -it -p 9432:9432 bblfsh/javascript-driver:v2.7.3 --log-level=debug
wget https://raw.githubusercontent.com/prestodb/presto/master/presto-main/src/main/resources/webapp/dist/query.js
go get -v -u gopkg.in/bblfsh/client-go.v3/cmd/bblfsh-cli

bblfsh-cli query.js| wc -l
 726827
```

PR #390, opened 2019-03-28
Set a default max message size for all gRPC clients and servers

dennwc:
Currently, clients call `DialOptions` to get the default dial options that should work well for a Babelfish gRPC client. But this config does not include the default size setting that is set for the server.

This change sets the default message size in both the client and the server options.

Related to https://github.com/bblfsh/sdk/issues/384

### DIFF
--- a/protocol/driver.go
+++ b/protocol/driver.go
@@ -20,32 +20,49 @@ import (
 
 //go:generate protoc --proto_path=$GOPATH/src:. --gogo_out=plugins=grpc:. ./driver.proto
 
+const (
+	mb = 1 << 20
+
+	// DefaultGRPCMaxMessageBytes is maximum msg size for gRPC.
+	DefaultGRPCMaxMessageBytes = 100 * mb
+)
+
 // ServerOptions returns a set of common options that should be used in bblfsh server.
 //
 // It automatically enables OpenTrace if a global tracer is set.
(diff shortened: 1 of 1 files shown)


protocol/driver.go
```diff
@@ -20,32 +20,49 @@ import (
 
 //go:generate protoc --proto_path=$GOPATH/src:. --gogo_out=plugins=grpc:. ./driver.proto
 
+const (
+	mb = 1024 * 1024
+
+	// DefaultGRPCMaxMsgSize is maximum msg size for gRPC (in bytes).
+	DefaultGRPCMaxMsgSize = 100 * mb
```
creachadair: (optional) I would probably just write this as `DefaultGRPCMaxMessageBytes = 100 << 20`.

protocol/driver.go
```diff
@@ -20,32 +20,49 @@ import (
 
 //go:generate protoc --proto_path=$GOPATH/src:. --gogo_out=plugins=grpc:. ./driver.proto
 
+const (
+	mb = 1024 * 1024
+
+	// DefaultGRPCMaxMsgSize is maximum msg size for gRPC (in bytes).
```
creachadair: Alternatively you can just put the units in the name: `DefaultGRPCMaxMessageBytes`

PR #392, opened 2019-04-03
Implement UTF-16 code point index

dennwc:
Some native drivers (like JS) use a UTF-16 code point positions.

This change improves existing UTF-8 positioning tests and implements a UTF-16 positional index.

### DIFF
--- a/uast/transformer/positioner/positions.go
+++ b/uast/transformer/positioner/positions.go
@@ -3,6 +3,7 @@ package positioner
 import (
 	"fmt"
 	"sort"
+	"unicode/utf16"
 	"unicode/utf8"
 
 	"gopkg.in/bblfsh/sdk.v2/uast"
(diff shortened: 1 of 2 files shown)


uast/transformer/positioner/positions_test.go
```diff
@@ -180,3 +194,55 @@ a3`
 		})
 	}
 }
+
+func TestPosIndexUTF16(t *testing.T) {
+	// Verify that a UTF-16 code point offset -> byte offset conversion works.
+	const source = `line1
```
creachadair: Please mention also in the comment that this case includes surrogate pairs, which are an important corner case for UTF-16.

uast/transformer/positioner/positions.go
```diff
@@ -107,12 +114,23 @@ func fromUnicodeOffset(idx *positionIndex, pos *uast.Position) error {
...
+}
+
 // runeSpan represents a sequence of UTF8 characters of the same size in bytes.
 type runeSpan struct {
-	firstRuneInd int // index of the first rune
-	byteOff      int // bytes offset of the first rune
-	numRunes     int // number of runes
-	runeSize     int // in bytes
+	firstRuneInd  int // index of the first rune
+	firstUTF16Ind int // index of the first UTF-16 code point
```
creachadair: Maybe I have missed something, but I find having these two separate is confusing: If there are no UTF-16 code points, how will this stay in sync? Or to put it another way, if someone requests the UTF-16 offset of a rune in a string with no UTF-16 code points, the handling of conversion _to_ surrogate form would need to be addressed. 

I think at least we should make it clear what the invariants are.

dennwc: Mentioned both offset and size invariants. Hope it helps.

dennwc:
@creachadair Ready for another round.

PR #393, opened 2019-04-04
Support UTF8 and UTF16 line-column index lookups

dennwc:
Some drivers only provide line-column information that with the column in either UTF8 runes or UTF16 code points.

This change adds necessary information to the index to calculate the byte offset from those line-column pairs.

As a consequence, spans are now added for each new line in addition to each sequence of equally-sized runes. This increases the memory required for the index from `O(1)` to `O(N)` (where `N` is the line count).

### DIFF
--- a/uast/transformer/positioner/positions.go
+++ b/uast/transformer/positioner/positions.go
@@ -128,17 +128,17 @@ type runeSpan struct {
 	// offset/index invariant:
 	// byteOff >= firstUTF16Ind >= firstRuneInd
 
-	firstRuneInd  int // index of the first rune
-	firstUTF16Ind int // index of the first UTF-16 code point
-	byteOff       int // bytes offset of the first rune
+	firstRuneInd  uint32 // index of the first rune
+	firstUTF16Ind uint32 // index of the first UTF-16 code point
+	byteOff       uint32 // bytes offset of the first rune
 
 	// size invariant:
 	// runeSize8 >= runeSize16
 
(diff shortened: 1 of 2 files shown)


dennwc:
I don't think it calculating both adds a lot to the complexity. And for UTF-16, current implementation uses character byte length calculated for UTF-8, so supporting both is free in this case. For UTF-8 we can, of course, save few bytes per span by getting rid of UTF-16 indexes, in the cost of code duplication.

creachadair:
> I don't think it calculating both adds a lot to the complexity. And for UTF-16, current implementation uses character byte length calculated for UTF-8, so supporting both is free in this case. For UTF-8 we can, of course, save few bytes per span by getting rid of UTF-16 indexes, in the cost of code duplication.

It's already expanded the types fairly substantially, to the point where this PR is attempting to save some bits by using shorter integer types. Moreover, it does a bunch of work eagerly that I think in most cases will not be needed. While separating them wouldn't necessarily reduce the algorithmic complexity—and might as you say increase the static code size a bit—it would be helpful to reduce the amount of work we do to eagerly compute offset indices for both representations ahead of time. 

Since we've just added UTF-16 support, it's a good time to consider those costs, I think.

dennwc:
Closing this. Turned out to be unrelated - the issue it was aimed to solve is related to non-UTF8 parse input. All drivers either use byte line-column or unicode (UTF8) offset.

PR #395, opened 2019-04-10
Use Go modules in drivers

dennwc:
This PR is the first step to Go modules. Starting from this version, drivers will use Go modules to manage dependencies, while SDK will continue to use `dep` for some time.

The new `bblfsh-sdk` tool is now able to update the driver from `dep` to modules automatically (tested on Go and Python drivers locally). Although note that it requires a new SDK to be tagged, or a hack to driver's `go.mod` file.

While on it, also implement tests in SDK that try to initialize a new driver, build it, run it and parse a test input. To make a new driver to actually "parse", PR changes a script in the driver skeleton to act like an "echo" driver. Thus, `bblfsh-sdk init` will now initialize a fully-functional driver.

There are also a few changes to the driver SDK version detection. It can now read both `go.mod` and `Gopkg.toml`.


build/sdk_update.go
```diff
@@ -310,3 +295,67 @@ func fixGitFolder(path string) string {
...
+		return err
+	}
+	if err := modTidy(root); err != nil {
+		return err
+	}
+	if err := modVendor(root); err != nil {
+		return err
+	}
+	// next, replace Go version in the driver's manifest
+	if err = updateGo(root, "1.12"); err != nil {
```
creachadair: Is this necessary? The only place this tag matters is if the Go standard library changes in incompatible ways, and it doesn't prevent earlier (or later) versions from building the module.

etc/skeleton/native/native.sh
```diff
@@ -1,3 +1,9 @@
-#!/usr/bin/env sh
-echo "TODO: native driver"
+#!/usr/bin/env bash
+while read -r req
```
creachadair: Can you please add a comment outlining briefly what this file does? E.g.,

```
# This implements a trivial native driver that returns its input unchanged
```

or words to that effect.

etc/skeleton/build.yml.tpl
```diff
@@ -12,9 +12,9 @@ native:
       - 'echo dependencies'
     run:
       - 'echo build'
-    artifacts:
-      - path: '/native/native-binary'
-        dest: 'native-binary'
+#    artifacts:
+#      - path: '/native/native-binary'
+#        dest: 'native-binary'
```
creachadair: Dead?

etc/build/Dockerfile.tpl
```diff
@@ -1,6 +1,11 @@
+# This file can be used directly with Docker.
+#
 # Prerequisites:
-#   dep ensure --vendor-only
+#   go mod vendor
 #   bblfsh-sdk release
+#
+# However, the preferred way is:
```
creachadair: Maybe say something about why this is preferred?

build/init.go
```diff
@@ -0,0 +1,91 @@
...
+		return err
+	}
+
+	if err := t.Execute(f, map[string]string{
+		"Language": language,
+	}); err != nil {
+		_ = f.Close()
+		return err
+	}
+	if err = f.Close(); err != nil {
```
creachadair:
```suggestion
	if err := f.Close(); err != nil {
```

build/init.go
```diff
@@ -0,0 +1,91 @@
...
+		return err
+	}
+
+	if err = UpdateSDK(root, opt.toUpdateOpt()); err != nil {
```
creachadair:
```suggestion
	if err := UpdateSDK(root, opt.toUpdateOpt()); err != nil {
```

build/git.go
```diff
@@ -36,3 +36,28 @@ func gitHasChanges(path string) (bool, error) {
...
 }
+
+func gitInit(path string) error {
+	if out, err := exec.Command("git", "init", path).CombinedOutput(); err != nil {
+		return fmt.Errorf("cannot init a repository: %v\n%s\n", err, string(out))
+	}
+	return nil
+}
+
+func gitAdd(root, file string) error {
```
creachadair: Maybe we could make this

```go
func gitAdd(root string, files ...string) error {
   git := exec.Command("git", append([]string{"add"}, files...)...)
   git.Dir = root
   …
   return nil
}
```

Then we wouldn't need a loop: `gitAdd(root, "Dockerfile", "go.mod", "go.sum")`

build/golang.go
```diff
@@ -31,3 +24,33 @@ func (d *Driver) depEnsure() error {
 	}
 	return nil
 }
+
+func goExec(dir string, args ...string) error {
```
creachadair: [optional] I sometimes find it helpful to have a command-builder:

```go
func goCmd(dir, modules string, args ...string) *exec.Cmd {
   cmd := exec.Command("go", args...)
   cmd.Stderr = os.Stderr
   cmd.Stdout = os.Stdout
   cmd.Dir = dir
   cmd.Env = append(os.Environ(), "GO111MODULE="+modules)
   return cmd
}

func goExec(dir string, args ...string) error { return goCmd(dir, "off", args...).Run() }

func goMod(dir string, args ...string) error { return goCmd(dir, "on", args...).Run() }
```

This also makes it easy to add additional bits to the command before running it, if necessary, without replicating the setup. YMMV.

build/golang.go
```diff
@@ -31,3 +24,33 @@ func (d *Driver) depEnsure() error {
...
+	cmd := exec.Command("go", args...)
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
+	cmd.Dir = dir
+	cmd.Env = append(os.Environ(), "GO111MODULE=on")
```
creachadair: Was this meant to be `=off` ?

build/golang.go
```diff
@@ -31,3 +24,33 @@ func (d *Driver) depEnsure() error {
...
+	args = append([]string{"mod"}, args...)
+	cmd := exec.Command("go", args...)
+	cmd.Stderr = os.Stderr
+	cmd.Dir = dir
+	cmd.Env = append(os.Environ(), "GO111MODULE=on")
+	return cmd.Run()
+}
+
+func modTidy(dir string) error {
+	return modExec(dir, "tidy")
```
creachadair: [optional, minor] Could be one-liners.

build/sdk_update.go
dennwc: This is not the change for `go.mod`, it's a change to the driver manifest that affects what version is used to build the driver. So yes, I think updating it to 1.12 matters.

etc/skeleton/native/native.sh
dennwc: Good catch, thanks! I definitely wanted to add a TODO for this.

etc/skeleton/build.yml.tpl
dennwc: Rather an example. Will add a comment as well.

build/git.go
dennwc: Considered this. Will do.

etc/skeleton/.travis.yml
```diff
@@ -1,20 +1,16 @@
...
 go:
-  - '1.11'
+  - '1.12.x'
 
 services:
   - docker
 
 env:
-  - BBLFSHD_VERSION=v2.11.8
+  - GO111MODULE=on BBLFSHD_VERSION=v2.12.1
```
bzz: If the intent is to have multiple env vars setup by default (rather than adding rows to the build matrix), I find this intent to be more clear when  [`global:` key is used in `env`](https://docs.travis-ci.com/user/environment-variables/#global-variables)

```yml
env:
  global:
    - GO111MODULE=on
    - BBLFSHD_VERSION=v2.12.1
```

build/git.go
```diff
@@ -52,3 +52,12 @@ func gitAdd(root, file string) error {
 	}
 	return nil
 }
+
+func gitRm(root, file string) error {
```
bzz: Same suggestion as in https://github.com/bblfsh/sdk/pull/395#discussion_r274486254 can be applied in here - using variadic function argument would allow to avoid [the loop](https://github.com/bblfsh/sdk/pull/395/commits/3f68b6466d97bb8bf6b7034d5549805354019186#diff-e8578c8c0c993fd45844c786dd815617R325).

etc/skeleton/README.md.tpl
```diff
@@ -15,16 +15,17 @@ Development Environment
...
 
 To initialize the build system execute: `go test ./driver`, at the root of the project. This will generate the `Dockerfile` for this driver.
 
-To execute the tests just execute `go run test.go`, this will execute the test over the native and the go components of the driver using Docker.
+To run the tests just execute `go run test.go`, this will start the test over the native and the Go components of the driver using Docker.
 
 The build is done executing `go run build.go`. To evaluate the result using a docker container, execute:
 `go run build.go test-driver && docker run -it test-driver`.
 
+If the project is located under `GOPATH`, run all the above with `GO111MODULE=on` environment variable,
```
bzz:
```suggestion
If the project is located under `$GOPATH`, run all the above with `GO111MODULE=on` environment variable,
```

etc/skeleton/README.md.tpl
```diff
@@ -15,16 +15,17 @@ Development Environment
...
+or move the project to any other directory outside of `GOPATH`.
```
bzz:
```suggestion
or move the project to any other directory outside of `$GOPATH`.
```

build/golang.go
dennwc: I'd love this to work, but a usual error text from `Run` is something like `"exit code 1"`. So it's necessary to catch `stderr` and replace an error text with it, for example.

But it seems like I was too lazy to catch it :man_facepalming:

dennwc: Nevermind, I will just redefine the `Run` method then.

driver/manifest/manifest.go
```diff
@@ -312,21 +320,61 @@ func Maintainers(open OpenFunc) ([]Maintainer, error) {
 	return list, nil
 }
 
-const sdkName = "gopkg.in/bblfsh/sdk"
+const (
+	sdkNameLegacy = "gopkg.in/bblfsh/sdk"
+	sdkName       = "github.com/bblfsh/sdk"
+)
 
 // maxVersionToml finds a maximal version of a Babelfish SDK in the list of the projects.
```
bzz: at this point I had to consult the internet on the difference between [maximum vs maximal](https://www.quora.com/What-are-the-differences-between-“maximum”-and-“maximal”-Especially-in-scientific-papers)

creachadair: In this case either is fine. I would have used "maximum" myself, but this is also correct.

etc/skeleton/.travis.yml
dennwc: Agree, this looks better :)

dennwc:
Ready for another round!

build/golang.go
creachadair: Ah: Note that the `Output` method handles this for you: If you call `c.Output()` the return value is the stdout, and stderr is captured for you: https://godoc.org/os/exec#ExitError

So if that's the main reason we're doing this, maybe this will help.

dennwc:
Still need to unbox that one, since it's [not printed](https://golang.org/src/os/exec/exec.go#L444) as a part of an error string. So the override for `Run` will still be there.

dennwc:
Done. Let's merge this, so the follow-up PR can be open for the review.

PR #396, opened 2019-04-11
UAST diff

dennwc:
Supersedes https://github.com/bblfsh/sdk/pull/309. See the original PR for details.


creachadair:
Before we merge this, we should go through and copy the important parts of the comment thread into the description here IMO.

dennwc:
Will add those comments as well. It still needs some work as well.

dennwc:
Rebased on the latest master, also cleaned up the code a bit. Ideally, need at least a few tests that check how the diff output looks like.

bzz:
Nice job!

Shall we also briefly talk/present only the API that this functionality exposes to the potential users and get their feedback early on, so there are less chances of the need for invasive chances later on?

I know that @Jan21 was using alternative solution for extracting this dataset of structural diffs for every commit https://github.com/src-d/datasets/tree/master/StructuralCommitFeatures so it might make sense to ask @creachadair at next TL meeting to ping ML team to take a quick look, WDYT?

dennwc:
PR doesn't expose or define any API yet. API is still TBD as well as if we will use this specific diff implementation or not.

PR #397, opened 2019-04-12
Support Go modules in SDK

dennwc:
This is the second step to finish the transition to Go modules.

PR bumps major SDK version to `v3`, switches all import paths from `gopkg.in` to Github, fixes docs, etc.

Also, the major version bump allows removing some deprecated symbols in packages. Similarly, take a chance to rename a `uast/yaml` (`uastyml`) package to `uast/uastyaml` (`uastyaml`) to match the package name and import path.


uast/uastyaml/uastyaml.go
```diff
@@ -0,0 +1,345 @@
+package uastyaml
```
creachadair: This is just the renamed file, right? (I wish GitHub distinguished moves better)

etc/skeleton/driver/normalizer/annotation.go
```diff
@@ -19,15 +18,6 @@ var Native = Transformers([][]Transformer{
 	},
 }...)
 
-// Code is a special block of transformations that are applied at the end
```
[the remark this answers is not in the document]
dennwc: It's already in the separate commit. I can split them out of the PR of course.

uast/uastyaml/uastyaml.go
dennwc: Correct. I also made temporary aliases in the old package.

etc/skeleton/driver/normalizer/annotation.go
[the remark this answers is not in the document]
creachadair: Ah, yeah sorry, I'm still used to squashing commits and we don't do that. Carry on. 😁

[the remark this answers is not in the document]
dennwc: I usually squash everything before opening a PR :)

dennwc:
@bzz Friendly ping

.travis.yml
```diff
@@ -8,19 +8,20 @@ go:
...
+env:
+  - GO111MODULE=on
 
 before_install:
-  - cd $GOPATH/src/gopkg.in/bblfsh/sdk.v2
+  - cd $GOPATH/src/github.com/bblfsh/sdk
   - make validate-commit
-  - wget -O dep https://github.com/golang/dep/releases/download/v0.4.1/dep-linux-amd64 && chmod +x dep
-  - ./dep ensure --vendor-only
+  - go mod download
```
bzz: a bit hard to understand how valuable is to download that before build/test  (e.g could be very valuable if it's warming up a cache for multiple runs but not sure if that is the case)

.travis.yml
```diff
@@ -8,19 +8,20 @@ go:
...
   fast_finish: true
   allow_failures:
     - go: tip
 
+env:
+  - GO111MODULE=on
 
 before_install:
-  - cd $GOPATH/src/gopkg.in/bblfsh/sdk.v2
+  - cd $GOPATH/src/github.com/bblfsh/sdk
```
bzz: not sure why is this still needed?

.travis.yml
dennwc: For now, it only separates the dependency download process from the test, nothing more. But if I'm not mistaken, there is a way to cache data across builds. So later we can add Go modules cache to it, for example.

.travis.yml
dennwc: In theory, it shouldn't matter anymore. But I would prefer to first make it work like this, and then start cutting up unneeded things. There may be some other scripts that rely on this behavior, for example.

etc/skeleton/driver/normalizer/annotation.go
```diff
@@ -19,15 +18,6 @@ var Native = Transformers([][]Transformer{
...
-// and can access original source code file. It can be used to improve or
-// fix positional information.
-//
-// https://godoc.org/gopkg.in/bblfsh/sdk.v2/uast/transformer/positioner
-var Code = []CodeTransformer{
```
bzz: If this means that on a update of the driver this needs to be removed - let's add it to the SDK release notes, so a driver maintainer is not confused. WDYT?

dennwc: Sure, will mention it. Drivers should already use a new version of this transform stage, so it's only a matter of removing it in the driver as well.

PR #399, opened 2019-05-07
Refactor the driver manifest (part 1)

dennwc:
Remove unused manifest fields and get runtime versions from the build manifest.

Previously the runtime and Go versions for the `README` were derived from the `manifest.toml`, while the actual driver build (in `Dockerfile`) was using versions from `build.yml`. This caused most drivers to be listed as requiring Go 1.12, but in fact, they were still built with Go 1.10. Same happened with native driver versions.

So instead of having random versions in `manifest.toml`, read them directly from `build.yml`.

Also, while on it, remove other unused fields in the manifest like OS and "information loss".

This change switches the `NativeVersion` type from `[]string` to `string`. It should not affect anything though, because they are no longer read from `manifest.toml` and it's unlikely that anyone except us was using those fields directly.

The change also includes a workaround for https://github.com/golang/go/issues/28065 to be able to use Go 1.12 in Alpine images.


build/build.go
```diff
@@ -56,6 +56,8 @@ type artifact struct {
 	Dest string `yaml:"dest"`
 }
 
+// should be in-sync with driver/manifest.buildManifest
```
creachadair: Is there a practical way for us to detect when they get out of sync? Even if we still have to fix regressions manually, it'd be nice if CI could fail to let us know that we need to.

driver/manifest/manifest.go
```diff
@@ -211,13 +180,58 @@ func Load(path string) (*Manifest, error) {
...
+		return m, err
+	}
 	m.Maintainers, err = Maintainers(open)
 	if err != nil {
 		return m, err
 	}
 	return m, nil
 }
 
+func extractVersion(s string) string {
```
creachadair: Please document what this is meant to do, e.g.,

```go
// extractVersion parses a semantic version from a Docker image name having
// the form base:label, where label is either a version or version-tag pair.
```

At least I think that's what's going on here. Maybe the name should indicate that too, say maybe `extractImageVersion`?

driver/manifest/manifest.go
```diff
@@ -211,13 +180,58 @@ func Load(path string) (*Manifest, error) {
...
+	if i := strings.LastIndexByte(s, ':'); i > 0 {
+		s = s[i+1:]
+	}
+	if i := strings.LastIndexByte(s, '-'); i > 0 {
+		s = s[:i]
+	}
+	return s
+}
+
+func LoadBuildManifest(m *Manifest, open OpenFunc) error {
```
creachadair: Please document the exported function. In particular, that it modifies its argument (which is a little odd for an exported API).

etc/build/Dockerfile.tpl
```diff
@@ -90,6 +94,9 @@ WORKDIR $DRIVER_REPO_PATH/
 
 ENV GO111MODULE=on GOFLAGS=-mod=vendor
 
+# workaround for https://github.com/golang/go/issues/28065
```
creachadair: Why is this one not also conditional, as in Stage 1.1 above?

driver/manifest/manifest.go
```diff
@@ -211,13 +180,58 @@ func Load(path string) (*Manifest, error) {
...
+	f, err := open("build.yml")
```
creachadair: Not for this change, but I think we should revisit this interface with OpenFunc. There are several cases in this file where we have to do careful fiddling with filehandles to make sure everything gets closed properly, when in all the cases I can see we just want the whole file anyway.

I think I might come back and do a separate no-op simplification pass after this gets merged, since this change is actually solving a user-facing problem and should take precedence.

etc/build/Dockerfile.tpl
dennwc: The previous workaround is inside the "native test" stage, so it applies only when the native language is Go. In this case, the stage is a "server build" which is always in Go.

build/build.go
dennwc: No, as long as both are unexported. But we can move the `buildManifest` to `manifest` package and this package will import it.

driver/manifest/manifest.go
dennwc: I think it's pretty common to return read-closers from similar API methods. But for this case you are right, it always loads small text files that fit into memory just fine.

We are not in a hurry with this change, so I can make it here as a separate commit.

build/build.go
dennwc: I have a better idea - it can be in the `internal` package, so it won't become our public API.

dennwc: Done

driver/manifest/manifest.go
dennwc: Done

build/build.go
```diff
@@ -13,15 +13,15 @@ import (
...
 	"github.com/bblfsh/sdk/v3/internal/docker"
-	"gopkg.in/yaml.v2"
 )
 
 const releaseManifest = ".manifest.release.toml"
 
 const (
 	dockerFileName = docker.FileName
-	manifestName   = "build.yml"
+	manifestName   = buildmanifest.Filename
```
creachadair: [optional nit] We seem to use `FileName` everywhere else, but `Filename` here. Separate packages, I realize, but it might help to be consistent.

internal/buildmanifest/manifest.go
```diff
@@ -0,0 +1,66 @@
+package buildmanifest
```
creachadair: Please include a [package comment](https://github.com/golang/go/wiki/CodeReviewComments#package-comments), `// Package buildmanifest …`

internal/buildmanifest/manifest.go
```diff
@@ -0,0 +1,66 @@
...
+
+import "gopkg.in/yaml.v2"
+
+const (
+	// Filename of a build manifest.
```
creachadair:
```suggestion
	// Filename of a build manifest relative to the package directory.
```

(assuming that is correct)

internal/buildmanifest/manifest.go
```diff
@@ -0,0 +1,66 @@
...
+	Filename = "build.yml"
+	// CurrentVersion os the manifest file format. See Manifest.SDK.
```
creachadair:
```suggestion
	// CurrentVersion is the manifest file format. See Manifest.SDK.
```

I would also suggest naming this `FileFormat` rather than "Version" since we have a lot of different version markers in play here (and `buildmanifest.FileFormat` reads pretty well at the usage site).

internal/buildmanifest/manifest.go
```diff
@@ -0,0 +1,66 @@
...
+
+// Artifact is a file copied from one path to another during the build.
+type Artifact struct {
+	Path string `yaml:"path"`
+	Dest string `yaml:"dest"`
+}
+
+// Manifest used for the declarative build system for drivers.
+type Manifest struct {
+	// SDK version. Used to track future changes to this file. Current version is '2'.
```
creachadair:
```suggestion
	// SDK version. Used to track future changes to this file. The current version is given by FileFormat.
```

But now I'm a little confused: Is this the SDK file format version or the SDK version? I think it's the first one (and hence my suggestion), but if that's correct the name `sdk` is a little odd. We might be stuck with the YAML name because of existing config files, but can we change the field name to reflect the semantics?

internal/buildmanifest/manifest.go
```diff
@@ -0,0 +1,66 @@
...
+	SDK string `yaml:"sdk"`
+
+	Native struct {
```
creachadair: What does `Native` mean in this context? Should this be `Driver`? (Again, I realize the YAML names may be fixed due to existing config files, but we control the struct layout)

internal/buildmanifest/manifest.go
```diff
@@ -0,0 +1,66 @@
...
+		// Image is a Docker image used as the native driver runtime.
+		Image string `yaml:"image"`
+		// Static is a list of files that will be copied from native driver's source directory
+		// to the final driver image. Note that those files cannot be modified by the build.
```
creachadair: Does "cannot" in this context mean "we will report an error" or "bad things will silently happen"?

If it's the first, we should (separately) make sure we check for it.

If it's the second, I suggest `must not be modified`, to emphasize it's a responsibility of the build tool rather than a restraint of the manifest user.

internal/buildmanifest/manifest.go
```diff
@@ -0,0 +1,66 @@
...
+		Static []Artifact `yaml:"static"`
+		// Deps is a list of apt/apk commands executed in the final driver image.
+		// This directive should be avoided since a different Docker image may be used instead of it.
+		Deps  []string `yaml:"deps"`
+		Build struct {
+			// Gopath directive sets a new gopath for the driver build. Only used by Go drivers.
```
creachadair: Can you please mention the format of this value? Is it a single directory, or a `a:b:c` style path list?

creachadair: (Even if it's just "whatever you would poke into `GOPATH`", it would be helpful to say that in the comment)

internal/buildmanifest/manifest.go
```diff
@@ -0,0 +1,66 @@
...
+			Gopath string `yaml:"gopath"`
+			// Image is a Docker image used to build the native driver.
```
creachadair: Is this `url:label` or `url` or either or other?

internal/buildmanifest/manifest.go
```diff
@@ -0,0 +1,66 @@
...
+			Image string `yaml:"image"`
+			// Deps is a list of shell commands to pull native driver dependencies.
+			// Note that those commands are executed before copying the driver files to the
+			// container, so they can be cached. See Run also.
+			Deps []string `yaml:"deps"`
```
creachadair: I find the names here somewhat confusing. "Deps" mostly makes sense—but "Add" seems to be the same idea that we called "Static" above, and "Run" seems like it should more properly be called "Build" (since it is the build instructions).

Also, the distinction between "Add" and "Artifacts" is a little obscured I think: Is it correct that "Add" artifacts are copied from the source tree to the build container, and "Artifact" artifacts are copied from the build container out into the deployment image?  I ask because I think for "Add" we really mean _container_ rather than _image_, whereas for deployment I think we really do mean _image_ (at least on the output side).

Anyway, I feel as if we could probably make the names and the comments a little more clear here.

internal/buildmanifest/manifest.go
```diff
@@ -0,0 +1,66 @@
...
+		} `yaml:"build"`
+		Test struct {
+			Deps []string `yaml:"deps"`
+			Run  []string `yaml:"run"`
+		} `yaml:"test"`
+	} `yaml:"native"`
+	Runtime struct {
+		// Version of Go used to build the driver server.
+		Version string `yaml:"version"`
+	} `yaml:"go-runtime"`
```
creachadair: I'm not sure I understand what we mean by "runtime" here. Does this mean "during the run of the build" (i.e., build time) or "during the run of the deployed container" (i.e., application runtime)? Maybe it doesn't matter, but it seems like we might be building with one version of Go but deploying into a container with another. Or if we really always want them to be the same, maybe the comment should say "…to build and run…".

internal/buildmanifest/manifest.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// Decode the manifest file.
```
creachadair:
```suggestion
// Decode updates m by decoding a YAML manifest from data.
```

It might also be worth noting that in case of error, the original contents of `m` are essentially garbage.

internal/buildmanifest/manifest.go
dennwc: The "driver" may have 3 different meaning in this context:
- Native driver (e.g. in Java).
- Driver server (in Go).
- The driver image (both).

So answering the question, this part of the structure is about "native" part only.

internal/buildmanifest/manifest.go
dennwc: Neither, it's more like "it's impossible to do so". I should probably use this wording in the comment.

build/build.go
dennwc: Unfortunately, the `Filename` is used in the original `manifest` package and other projects like `bblfshd` use that constant. So it cannot be renamed this way without breaking the build.

internal/buildmanifest/manifest.go
dennwc: It's an SDK version that first introduced the format. I changed the comment, hope it's better now.

internal/buildmanifest/manifest.go
dennwc: Clarified comments a bit. Is it any better?

The `Add` field is similar to Docker `ADD` directive. At first `Static` was named `Add` as well, but it was confusing, so I changed it to emphasize that those files cannot be changed by the build and are more like static assets.

internal/buildmanifest/manifest.go
dennwc: Added few more comments here.

dennwc:
Ready for another round.

build/build.go
creachadair: Ah, well. It'll be fine as is then.

internal/buildmanifest/manifest.go
creachadair: I think both qualify as static assets—in each case the specified files are copied verbatim. The comments do help, but I feel like there isn't a clear distinction such that these should have different names. The fact that one uses the Docker `ADD` mechanism doesn't really matter when writing the config, right?

I guess I was hoping we could settle on one term for static assets and use that throughout the config, assuming I didn't overlook some important distinction.

creachadair: Perhaps more importantly, it'd be nice if we can harmonize the terminology for the different categories of static assets: In the one case, for the build container, in the other for the application image. 

Maybe the name of the target? E.g., `BuildAssets`, `ImageAssets`? I'm open to other ideas too of course. (I know it may seem silly to worry so much about names, but names carry a lot of the weight of the documentation)

dennwc: I'm all in for making the docs as clear as possible. In other cases it will probably be left untouched for months to come :grin:

Regarding `ADD`, indeed it doesn't matter but it may provide a clue for someone familiar with Docker. But `Static` won't.

The only specific field is `Artifacts` - it copies files from build container to the image instead of from source as other similar directives.

`BuildAssets` and `ImageAssets` (or maybe `*Files`?) sounds good to me.

We might also increment the format version to reflect this change to config as well, but it's probably doesn't make that much sense, given an effort that is required.

dennwc:
Renamed fields to `*Assets`, ready for another round.

dennwc:
Will wait for @bzz comments as well. Working on part 2 (aliases) in the meantime.

Issue #382, opened 2019-03-20
feature request: provide Unicode positions


vmarkovtsev:
As we have recently found out, the positions of UAST nodes must be measured in bytes, not in runes. However, we require working with strings, so we had to [build the conversion ourselves](https://github.com/src-d/lookout-sdk-ml/blob/master/lookout/core/bytes_to_unicode_converter.py). This code is not straightforward.
Besides, all kinds of crazy things may happen, e.g. the [line number can change](https://github.com/src-d/lookout-sdk-ml/pull/75).

I suggest adding the function to recalculate positions to Babelfish, either as an API extension or to the clients. It is not cool to calculate them by hand, and they are different from IDE (expected) positions anyway.

bzz:
Not sure how generally applicable is that across all the drivers, but I know some work has been done on it e.g on js driver, in order to make this happen https://github.com/bblfsh/javascript-driver/pull/51

dennwc:
This was a reverse operation: Babelfish expects bytes positions, but JS was writing Unicode positions.

The feature request makes total sense, but I think we can't add Unicode positions as a field - it will make UAST files even larger.

So I propose to send the whole source file as a part of UAST. Then we can recalculate Unicode positions and even tokens on demand.

vmarkovtsev:
It is also possible to add a parameter to the parse request that the client wants Unicode positions. Normally you don't need both at the same time.

dennwc:
This makes sense. We can extend the parsing protocol with one more option for this. Still, the `Position` node will need some flag to signal that it's stored as Unicode position.

dennwc:
As per discussion [on Slack](https://src-d.slack.com/archives/C7UDG9VNY/p1553103406102100) (non-public link), this sounds like a "wontfix" for the SDK:

> @creachadair: The representation of position is tricky: Even languages that use Unicode internally do not necessarily agree on the distinction between codepoint offset and character. For different normalizations you may well get different offsets, and whether or not the normalization happens implicitly varies.
>
> Byte offset is the only reliable _universal_ baseline, and so the real question IMO is how to get codepoint offsets into the API. Personally, I think it makes sense to put the code for transforming a (bytes, byte-offset) pair into a codepoint-offset, line number, column offset, etc., in a common location. But I do _not_ think it makes sense to plumb a bunch of switches through the APIs for this.

[...]

PR #401, opened 2019-05-23
Expose positional index to use in libuast

dennwc:
https://github.com/bblfsh/sdk/issues/382 requires exposing the positional index to be used in `libuast`/clients, so this change makes it public. Not sure why we had it private in the first place.

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/401)


creachadair:
Unexported is a good default. :smile_cat:

PR #406, opened 2019-06-06
Implement node children iterator

dennwc:
This issue came up multiple times recently. Some client functions want to list all children nodes exactly one level deep. This code is not a one-liner since it requires the user to check the type of the node and indirect, if necessary. It also requires to flatten arrays for object fields.

Thus, it seems we need to add some helper to expose this functionality to clients. The best way, in my opinion, is to make a new iteration order.

This PR adds a new `Children` iteration order that does all required checks and indirections.

Also, while here, add documentation for types and methods in the package.

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/406)

### DIFF
--- a/uast/nodes/iter.go
+++ b/uast/nodes/iter.go
@@ -2,6 +2,7 @@ package nodes
 
 import "fmt"
 
+// Iterator over nodes.
 type Iterator interface {
 	// Next advances an iterator.
 	Next() bool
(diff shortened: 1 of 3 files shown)


dennwc:
@bzz Friendly ping.

dennwc:
Merging, since it hits the review deadline of one week. @bzz let me know if you have any comments on this one.

bzz:
Sorry for somehow letting it slip - this is a great improvement!
I was just thinking what would it take to surface it in all the clients..

PR #412, opened 2019-07-15
docs: Fix godoc.org links in exported name comments.

creachadair:
I found these while investigating https://github.com/bblfsh/go-driver/issues/48. We may also need to update the generated code itself, but for now this is only a comment fix with no functional changes.

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/412)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,7 @@ validate-commit: bindata
 	if  [ $$? -eq 0 ] ; then \
 		git diff|cat; \
 		echo >&2 "generated bindata is out of sync"; \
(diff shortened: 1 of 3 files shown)


creachadair:
Ah, I see what I've done wrong here. I'll re-sync and update the error message to include fix instructions.

bzz:
It's impressive and puzzling that this change seems to also affect the code coverage

creachadair:
Code coverage tools are like the boy who cried wolf—their sensitivity is out of proportion to the value of their signal. :smile:

PR #432, opened 2019-09-12
Propagate more error types through gRPC

dennwc:
Currently, our gRPC protocol only uses gRPC error codes for propagation. This significantly limits the number of possible errors that we can distinguish on the client side. This may be a problem for the client that want to skip some error cases like "unknown language" and "invalid file encoding".

This PR defines a new `ErrorDetails` protobuf message that will now be attached to all gRPC errors. It will provide all the information the client needs to reconstruct the error passed through the RPC channel.

In case the information is not available, implementation will fallback to the old behavior and will use error codes to distinguish errors.

The change also moves some of the error types from `bblfshd` to SDK. Once the new SDK version is released, we can remove those error types from `bblfshd` packages.

It also adds tests for gRPC error propagation.

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/432)


ncordon:
Can you open an issue in the `bbfshd` to remind us to delete the error types once the SDK is released?

ncordon:
___
*[driver/driver.go, line 38 at r2](https://reviewable.io/reviews/bblfsh/sdk/432#-LofcXvW0QZJt4t2s6TE:-LofcXvW0QZJt4t2s6TF:b-vd8sd3) ([raw file](https://github.com/bblfsh/sdk/blob/5d86609aaeaf7645367a985e0f8eb00d490e5ecf/driver/driver.go#L38)):*
> ```Go
> 
> // ErrMissingDriver indicates that a driver image for the given language
> // can not be found.
> ```

This is ultra peaky, but *cannot* instead of *can not*

ncordon:
___
*[protocol/driver_test.go, line 69 at r2](https://reviewable.io/reviews/bblfsh/sdk/432#-LoffK5iACJJqHOX4a8-:-LoffK5iACJJqHOX4a80:b4tvk09) ([raw file](https://github.com/bblfsh/sdk/blob/5d86609aaeaf7645367a985e0f8eb00d490e5ecf/protocol/driver_test.go#L69)):*
> ```Go
> 	var cases = []driverMock{
> 		{name: "success", uast: defaultUAST()},
> 		{name: "partial parse", uast: defaultUAST(), err: driver.ErrSyntax.Wrap(errors.New("invalid source"))},
> ```

What is a partial parse?

PR #434, opened 2019-09-13
Splits suffixes in comments correctly

### LINKED ISSUE #433: Comments including utf8 characters are not parsed correctly
Go client:
```go
package main

import (
	"fmt"
	"io/ioutil"
	"net/http"

	bblfsh "github.com/bblfsh/go-client/v4"
	"github.com/bblfsh/go-client/v4/tools"
)

const fileURL = "https://github.com/pytorch/tutorials/raw/master/advanced_source/neural_style_tutorial.py"

func main() {
	httpRes, err := http.Get(fileURL)
	if err != nil {
		panic(err)
	}
	if httpRes.StatusCode != http.StatusOK {
		panic("not 200 status")
	}
	content, err := ioutil.ReadAll(httpRes.Body)
	if err != nil {
		panic(err)
	}

	client, err := bblfsh.NewClient("localhost:9432")
	if err != nil {
		panic(err)
	}

	res, _, err := client.NewParseRequest().Language("python").Content(string(content)).UAST()
	if err != nil {
		panic(err)
	}

	it, _ := tools.Filter(res, "//uast:FunctionGroup")
	for it.Next() {
		fmt.Println("node")
	}
}

```

works without any errors.

Python client:
```python
import bblfsh
import requests

FILE_URL = "https://github.com/pytorch/tutorials/raw/master/advanced_source/neural_style_tutorial.py"


def main():
    content = requests.get(FILE_URL).text

    client = bblfsh.BblfshClient("localhost:9432")
    ctx = client.parse(FILE_URL.split("/")[-1], None, content)
    it = ctx.filter("//uast:FunctionGroup")
    for node in it:
        print("node")


main()
```

results in:
```
node
node
node
node
node
node
node
node
node
node
node
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte

The above exception was the direct cause of the following exception:

SystemError: <class 'UnicodeDecodeError'> returned a result with an error set

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "reproduce.py", line 17, in <module>
    main()
  File "reproduce.py", line 13, in main
    for node in it:
  File "/Users/smacker/tmp/bblfsh_debug/venv/lib/python3.6/site-packages/bblfsh/node_iterator.py", line 29, in __next__
    self._last_node = Node(node_ext=next_node, ctx=self.ctx)
  File "/Users/smacker/tmp/bblfsh_debug/venv/lib/python3.6/site-packages/bblfsh/node.py", line 141, in __init__
    self.internal_node = node_ext.load()
SystemError: <built-in method load of pyuast.NodeExt object at 0x1039a47b0> returned a result with an error set
```

ncordon:
The suffixes of the comments in semantic tree were not split correctly, due to slices of strings working with bytes in Go, and not with unicode characters. For example, without this fix, in:

```
// “comment”
```

the final `”` was translated into `b'\xe2\x80\x9d'`, but the suffix was tell to cut the slice of bytes in the position `i`, where `i` is the `\xe2`, not the final of the unicode character (which would be `x9d`). And that's because a unicode character can span more than one byte.

Closes #433 

Signed-off-by: ncordon <nacho.cordon.castillo@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/434)


uast/transformer/semantic.go
```diff
@@ -194,8 +195,10 @@ func (c *commentElems) Split(text string) bool {
 
 	c.Suffix = ""
 	if i >= 0 {
-		c.Suffix = text[i+1:]
-		text = text[:i+1]
+		// compute the unicode length (could be more than just one byte)
+		_, ulen := utf8.DecodeRune([]byte(text[i:]))
```
dennwc: This works exactly for 1 Unicode rune at the end of the string. You'll need to loop here to decode all runes.

Since the same bug is probably there for prefixes as well, I think it may be better to convert string to runes first (`[]rune(s)`) and then apply other logic to runes instead of bytes.

ncordon: I am going to answer the second part first. I have thought about that, and with prefixes there are no problems (I think), because in the case of `Prefix`es:

```
	i := strings.IndexFunc(text, func(r rune) bool {
		return !c.isTab(r)
	})

	c.Prefix = ""

	if i >= 0 {
		c.Prefix = text[:i]
		text = text[i:]
	}
```

Note that `i` is actually the position (in bytes of the first byte of the rune), so for prefixes it would work fine, because we take until that first byte and then the rest of the string.

To the first observation you made: `i` is the first byte position for the last thing that is not a tab. The problem taking suffixes is that I need to include that character into the Suffix, so taking from the next byte onwards would just simply not work when the unicode character contains more than one byte. I have changed the test to include more utf8 (with more than one byte) characters at the end of the string. It works :smile:

ncordon: Also, yes, it would better to apply the logic to `[]rune` instead of `[] byte`. But we do not have methods to work with `[] rune` (prefixes, suffixes, trim and such) and a lot to work with strings. Also if we convert to `[]rune` we would have to do it at the beginning and stick with `[] rune` to avoid a lot of conversions back and forth from `strings`. For now this fix works well, and I would not want to touch anything else in the method really.

ncordon:
Sorry @dennwc, I was gathering strength to change the whole method to runes. Even if it is not so urgent it can be a good exercise for me to become more fluent in Go. So I am going to try doing it :smiley_cat:

ncordon:
___
*[uast/transformer/semantic.go, line 199 at r1](https://reviewable.io/reviews/bblfsh/sdk/434#-LofbTIdfT8qwzR9vNKZ-r1-199:-Lq6gY5V83yPBMir23i9:bjw7oxq) ([raw file](https://github.com/bblfsh/sdk/blob/897b0d4331149d980477e8b236c3014bdf635692/uast/transformer/semantic.go#L199)):*

@ncordon Sorry, it was my mistake. I see now why this specific change works. The only suggestion would be to change `DecodeRune` to `DecodeRuneInString` to avoid conversion to `[]byte`.

Also, since you feel uncomfortable with changing the code to `[]rune`, please add a `TODO` here and open an issue in SDK. I think we still need to address the same issue for the prefix (if it contains other Unicode space-like characters).

I left a few notes here to point places relevant to an issue with prefixes. Those are just for the context, no action is needed for this PR.
</blockquote></details>

@kuba-- I will test it once I have it all changed to runes.

uast/transformer/semantic.go
```diff
@@ -161,96 +164,102 @@ func (c *commentElems) isTab(r rune) bool {
...
+}
+
+// Returns the first index i of runes satisfying f(runes[i]), len(runes) otherwise
+func firstIndexFunc(runes []rune, f func(r rune) bool) int {
+	for i, r := range runes {
+		if (f(r)) {
+			return i;
+		}
 	}
+	return len(runes);
```
kuba--: maybe `return -1`, how do you know if it was found? you will have to compare to `len`

kuba--: Btw. you can try to reuse a func. from std. library:
https://golang.org/src/strings/strings.go?s=18931:18982#L758

uast/transformer/semantic.go
```diff
@@ -161,96 +164,102 @@ func (c *commentElems) isTab(r rune) bool {
...
+}
 
-	if !strings.HasPrefix(text, c.StartToken) || !strings.HasSuffix(text, c.EndToken) {
-		return false
+// Returns the last index i of runes satisfying f(runes[i]), -1 otherwise
+func lastIndexFunc(runes []rune, f func(r rune) bool) int {
+	for i := range runes {
```
kuba--: Check std. lib. 
I think it's better to do:
```
for i := len(s); i > 0; {
		r, size := utf8.DecodeLastRuneInString(s[0:i])
		i -= size
...
```

uast/transformer/semantic.go
```diff
@@ -161,96 +164,102 @@ func (c *commentElems) isTab(r rune) bool {
...
+	return -1;
+}
 
-	// find prefix (if not already trimmed)
-	i := strings.IndexFunc(text, func(r rune) bool {
-		return !c.isTab(r)
-	})
+func (c *commentElems) findPrefix(f func(r rune) bool) {
+	runes := []rune(c.Text)
+	i := firstIndexFunc(runes, f)
```
kuba--: so maybe these functions should return a substring instead of index

uast/transformer/semantic.go
```diff
@@ -161,96 +164,102 @@ func (c *commentElems) isTab(r rune) bool {
...
-	} else {
-		c.Prefix = text
-		text = ""
+func commonPrefix(a []rune, b []rune) []rune {
+	r := []rune{}
+	for i := 0; i < len(a) && i < len(b); i++ {
+		if (a[i] != b[i]) {
+			break;
+		} else {
+			r = append(r, a[i])
```
kuba--: do we need to create a new slice? Cannot we just return `a[:i+1]`

kuba--:
@ncordon - check CI

ncordon:
> @ncordon - check CI

The CI is failing because of the API limit @kuba--

uast/transformer/semantic.go
ncordon: Changed to avoid creation of extra slice!

uast/transformer/semantic.go
ncordon: Well, `lastIndexFunc` returns `len(runes)` for 2 reasons:
1. That way I avoid `if-elses` to check slice the string (at least in the use cases for this file)
2. It seems more consistent to me: `firstIndexFunc` loops from left to right and returns either the first index verifying the predicate or the first index out of bounds (`len(runes)`). `lastIndexFunc` does the same but from right to left, so out of bounds would be `-1`. Again this makes possible to avoid a lot of `if-elses` in the code.

The `std` version was used before, but we agreed on making the whole code work with `[] rune` instead of strings (the `std` version takes strings as argument). It can still be used of course, but it implies more conversions `[] rune <-> string`

uast/transformer/semantic.go
ncordon: If all of you agree it is better to use the `std` version I can change it, but again this would imply more conversions `string <-> [] rune`

uast/transformer/semantic.go
ncordon: You mean `firstIndexFunc` and `lastIndexFunc`?

uast/transformer/semantic.go
kuba--: It's ok to use runes, my point was to take a look into std. lib, because for instance returning len(s) if you don't find a rune is kind of not intuitive (in most cases you expect -1). On the other hand I know returning a len can be handy, because you can later call
`s[:index()]`, but again it's a special case, so maybe we should just return a substring/slice instead of index.

ncordon: :thinking: It's very likely I am not having into account Go conventions. So as you wish, I can change it to `-1`s

uast/transformer/semantic.go
kuba--: both

ncordon: The downside of that is that I would have to call `TrimPrefix` or `TrimSuffix` inside `findPrefix` or `findSuffix`. And `firstIndexFunc` is also used to compute only the prefix (as in the tab calculation inside the loop), so making them return `(prefix, rest of the text)` or `(text minus suffix, suffix)` is not suitable either. Also it should not return strings IMHO because if it is used to compute the tab, therefore it needs runes. If it returns strings I am adding another conversion `string <-> [] rune` again

ncordon:
___
*[coverage.txt, line 2 at r4](https://reviewable.io/reviews/bblfsh/sdk/434#-LqBw8h0E5dcf6avLutL:-LqC9NnzAxrpaz-RZxwy:b-896fja) ([raw file](https://github.com/bblfsh/sdk/blob/6970545b09cf1ffbf94997ae1a97005f4fea0c12/coverage.txt#L2)):*

Please delete this file. It's probably from a code coverage tool.
</blockquote></details>

Done!

ncordon:
___
*[go.mod, line 24 at r4](https://reviewable.io/reviews/bblfsh/sdk/434#-LqC6u0S8N0oVsrSL4hG:-LqC9QHv0dGb2MeZPV9Q:bez4u1g) ([raw file](https://github.com/bblfsh/sdk/blob/6970545b09cf1ffbf94997ae1a97005f4fea0c12/go.mod#L24)):*

Unrelated change. Is it because you used pprof for debugging, or it's from `go mod tidy` with no other changes?
</blockquote></details>

I have not used this. Probably it has been `go mod tidy`

ncordon:
___
*[uast/transformer/semantic_test.go, line 44 at r4](https://reviewable.io/reviews/bblfsh/sdk/434#-LqC6WJC5Hc3stpFh5yG:-LqCCU-ABhgBxGSHkQ4m:blk8udn) ([raw file](https://github.com/bblfsh/sdk/blob/6970545b09cf1ffbf94997ae1a97005f4fea0c12/uast/transformer/semantic_test.go#L44)):*

`"/*\t\t\u0020 ½ */"`? Also, isn't `\u0020` a space?
</blockquote></details>

Changed!. Yes, it is, I took it from the `unicode.IsSpace`documentation

ncordon:
___
*[uast/transformer/semantic.go, line 149 at r2](https://reviewable.io/reviews/bblfsh/sdk/434#-Lok9QDM0ZrYDd6VeA89:-LqCCkxF0g5qKzp1R01p:bx3iq6m) ([raw file](https://github.com/bblfsh/sdk/blob/7b9296e3460adc6949e7a2fba23a4bf2f13ca998/uast/transformer/semantic.go#L149)):*

Related to the prefix issue: here we accept any Unicode character of class "space" to be a part of the prefix, so we can match some multi-byte ones as well.
</blockquote></details>

Addressed

ncordon:
___
*[uast/transformer/semantic.go, line 181 at r3](https://reviewable.io/reviews/bblfsh/sdk/434#-LofbTIdfT8qwzR9vNKZ-r3-181:-LqCCnb0AVdkbBNHiA3k:b-1gz30f) ([raw file](https://github.com/bblfsh/sdk/blob/10186598370d9676b4e2100cce0cd951db78c81c/uast/transformer/semantic.go#L181)):*

Yes, if you check `strings.Index`, it will return `-1` if the index was not found. Also please remove `;`.
</blockquote></details>

Changed, but I am not comfortable with this change. It obscures code IMO

ncordon:
___
*[uast/transformer/semantic.go, line 181 at r3](https://reviewable.io/reviews/bblfsh/sdk/434#-LofbTIdfT8qwzR9vNKZ-r3-181:-LqCCxy667BZnCNooUgv:bdxiwlb) ([raw file](https://github.com/bblfsh/sdk/blob/10186598370d9676b4e2100cce0cd951db78c81c/uast/transformer/semantic.go#L181)):*
<details><summary><i>Previously, ncordon (Nacho Cordón) wrote…</i></summary><blockquote>

Changed, but I am not comfortable with this change. It obscures code IMO
</blockquote></details>

The -1 thing I mean, not the `;`

ncordon:
___
*[uast/transformer/semantic.go, line 186 at r3](https://reviewable.io/reviews/bblfsh/sdk/434#-LofbTIdfT8qwzR9vNKZ-r3-186:-LqCDC13-fmPiea2Lwsp:b6ehy40) ([raw file](https://github.com/bblfsh/sdk/blob/10186598370d9676b4e2100cce0cd951db78c81c/uast/transformer/semantic.go#L186)):*

@ncordon you are right in the sense that we will end up converting back and forth between `string` and `[]rune`. Unfortunately there is no helper in stdlib that works on an array of runes.

But, I also see that the result of the `firstIndexFunc` and `lastIndexFunc` function is used to split the string, not an array of runes. So @kuba-- has a point here: instead of returning a rune offset, you can calculate the byte offset in the string, and then split the string using it.
</blockquote></details>

Please note that `firstIndexFunc` is also used to compute the tab. And that is made in `[] rune` not bytes

ncordon:
___
*[uast/transformer/semantic.go, line 188 at r4](https://reviewable.io/reviews/bblfsh/sdk/434#-LqC-DPx7pyG4W3cp9rP:-LqCDLjlBIPffcx_8NCf:b-896fja) ([raw file](https://github.com/bblfsh/sdk/blob/6970545b09cf1ffbf94997ae1a97005f4fea0c12/uast/transformer/semantic.go#L188)):*

Please remove parentheses.
</blockquote></details>

Done!

ncordon:
___
*[uast/transformer/semantic.go, line 220 at r5](https://reviewable.io/reviews/bblfsh/sdk/434#-LqFyYQH9ou5mW-SipCt:-LqGUbFWB6Wzschk56j0:b-896fja) ([raw file](https://github.com/bblfsh/sdk/blob/9f978b316c0bcbcde5975fd2a7b6c24b48487bb2/uast/transformer/semantic.go#L220)):*

Please remove parentheses.
</blockquote></details>

Done!

ncordon:
___
*[uast/transformer/semantic.go, line 197 at r3](https://reviewable.io/reviews/bblfsh/sdk/434#-LofbTIdfT8qwzR9vNKZ-r3-197:-LqGUd--46SGs0F00znA:b-896fja) ([raw file](https://github.com/bblfsh/sdk/blob/10186598370d9676b4e2100cce0cd951db78c81c/uast/transformer/semantic.go#L197)):*

Or index in bytes, as @kuba-- mentioned in the previous comment. @ncordon should address  your concern, right?
</blockquote></details>

Done!

ncordon:
___
*[uast/transformer/semantic.go, line 186 at r3](https://reviewable.io/reviews/bblfsh/sdk/434#-LofbTIdfT8qwzR9vNKZ-r3-186:-LqGUfAN01r2Eh18fcxG:bbt64hl) ([raw file](https://github.com/bblfsh/sdk/blob/10186598370d9676b4e2100cce0cd951db78c81c/uast/transformer/semantic.go#L186)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

My intention was  - `firstIndexFunc` as an independent thingy should follow some standard and return -1. BUT in our case it's used in one place and having `-1` as an output is not handy. So what I suggested in comment(s) bellow is to return a `[]rune` (instead of an index).
In other words, having
```
func firstSliceFunc(runes []rune, f func(r rune) bool) []rune {
        var i int
	for r := range runes {
		if (f(r)) {
                       return runes[:i+1]
		}
               i++
	}
	return runes[:i]
``` 

So, in a use case instead of doing:
```
i := firstIndex
s = s[:i]
```
you can just:
```
s = firstSliceFunc(s, f)
```

and similar approach for `last...`
Do you know what I mean?
</blockquote></details>

Changed now! Thanks, I was not understading correctly what you meant

ncordon:
___
*[uast/transformer/semantic.go, line 242 at r2](https://reviewable.io/reviews/bblfsh/sdk/434#-Lok9g_I6vYUE0hVDZD-:-LqGUq3c8G2ugw_KtzKd:bptnfit) ([raw file](https://github.com/bblfsh/sdk/blob/7b9296e3460adc6949e7a2fba23a4bf2f13ca998/uast/transformer/semantic.go#L242)):*

Related to the prefix issue: here we check the `tab` which may contain multi-byte characters, but we compare bytes instead of runes.
</blockquote></details>

Addressed with the changes to `[]rune`

ncordon:
___
*[uast/transformer/semantic_test.go, line 44 at r4](https://reviewable.io/reviews/bblfsh/sdk/434#-LqC6WJC5Hc3stpFh5yG:-LqGUubXBfKPVFacDQEe:b-jflwij) ([raw file](https://github.com/bblfsh/sdk/blob/6970545b09cf1ffbf94997ae1a97005f4fea0c12/uast/transformer/semantic_test.go#L44)):*

I mean, the `0x20` is an ordinary space. You probably wanted something else, like any other kind of Unicode space, correct?
</blockquote></details>

Changed now

ncordon:
___
*[uast/transformer/semantic.go, line 177 at r5](https://reviewable.io/reviews/bblfsh/sdk/434#-LqFyww5E8qYFMQm5663:-LqGV0-c-mxRjLJ4e9PW:b-896fja) ([raw file](https://github.com/bblfsh/sdk/blob/9f978b316c0bcbcde5975fd2a7b6c24b48487bb2/uast/transformer/semantic.go#L177)):*

Please remove parentheses.
</blockquote></details>

Done!

ncordon:
___
*[go.mod, line 24 at r4](https://reviewable.io/reviews/bblfsh/sdk/434#-LqC6u0S8N0oVsrSL4hG:-LqGV1pK17ulgBuklnCe:b3w1ejf) ([raw file](https://github.com/bblfsh/sdk/blob/6970545b09cf1ffbf94997ae1a97005f4fea0c12/go.mod#L24)):*

Can you please also check that you don't have any temporary files? `go mod tidy` processes all files in the directory, not only ones committed to Git.
</blockquote></details>

Done! Sorry, I do not know what happened there. I was using `make test-coverage` which generated the `coverage.txt`, but once I deleted it and did `go mod tidy` again these things disappeared

uast/transformer/semantic.go
```diff
@@ -161,96 +164,106 @@ func (c *commentElems) isTab(r rune) bool {
...
-	if c.DoTrim {
-		text = strings.TrimLeftFunc(text, unicode.IsSpace)
+func max(x, y int) int {
+    if x > y {
+        return x
+    }
+    return y
+}
+
+// Returns the first index i of runes satisfying f(runes[i]), -1 otherwise
```
kuba--: update or remove the comment

ncordon: Done!

ncordon:
___
*[driver/manifest/discovery/discovery_test.go, line 41 at r8](https://reviewable.io/reviews/bblfsh/sdk/434#-LqkA8QvEK1EdYKpmGmT:-Lqv5-J66gjfaF88VEsa:b-896fja) ([raw file](https://github.com/bblfsh/sdk/blob/481de6b05d4a247f5dafcb19a2905649452dbaa5/driver/manifest/discovery/discovery_test.go#L41)):*

Can you please rebase? We made a proper fix for this.
</blockquote></details>

Done!

ncordon:
___
*[uast/transformer/semantic.go, line 147 at r8](https://reviewable.io/reviews/bblfsh/sdk/434#-LqkDVAfBWxbi4k6Q0nQ:-Lqv5qKzDHNj7U5Enx3Z:b-896fja) ([raw file](https://github.com/bblfsh/sdk/blob/481de6b05d4a247f5dafcb19a2905649452dbaa5/uast/transformer/semantic.go#L147)):*

In Go, comments must begin with a function name: `isTabToken check if a token is a tab. A tab is defined ...`.

Same for other comments.
</blockquote></details>

Done!

ncordon:
___
*[uast/transformer/semantic.go, line 189 at r8](https://reviewable.io/reviews/bblfsh/sdk/434#-LqkDrz6DWkayHM3lTT3:-Lqv6HwF5JvRkEraGF92:b-896fja) ([raw file](https://github.com/bblfsh/sdk/blob/481de6b05d4a247f5dafcb19a2905649452dbaa5/uast/transformer/semantic.go#L189)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

For sure.. - I suggest to `for i := len(runes) - 1; i > 0; i--`
</blockquote></details>

Done!

ncordon:
___
*[uast/transformer/semantic.go, line 247 at r8](https://reviewable.io/reviews/bblfsh/sdk/434#-LqkBmvUEno8ylPsvnaW:-Lqv6T002PTodLZvnk_k:b-896fja) ([raw file](https://github.com/bblfsh/sdk/blob/481de6b05d4a247f5dafcb19a2905649452dbaa5/uast/transformer/semantic.go#L247)):*

```golang
var tab []rune
```
</blockquote></details>

Done!

ncordon:
___
*[uast/transformer/semantic.go, line 198 at r8](https://reviewable.io/reviews/bblfsh/sdk/434#-LqkCngn2L1Mpw95Zuti:-LqvqEPV7zepqrL2ToIv:b-s666ez) ([raw file](https://github.com/bblfsh/sdk/blob/481de6b05d4a247f5dafcb19a2905649452dbaa5/uast/transformer/semantic.go#L198)):*

Can we reuse it? Both `findPrefix` and `findSuffix` convert the whole string this way.

Either consider using `utf8.DecodeRuneInString`/`utf8.DecodeLastRuneInString` or somehow share the slice that is converted to runes.
</blockquote></details>

I have changed it so that `findPrefix` and `findSuffix` reuse it and also the main body of the function, but certainly I would have preferred to keep the former version because of readability of the main algorithm of `Split`. We can do what you consider best, but my opinion is that what you save from those 3 conversions is not worth obscuring the main body of the `Split` method, and the algorithm is still linear in the length of the string

ncordon:
___
*[driver/manifest/discovery/discovery_test.go, line 41 at r8](https://reviewable.io/reviews/bblfsh/sdk/434#-LqkA8QvEK1EdYKpmGmT:-LrJhq_cDoTWnUA-xEtK:b-33bzmm) ([raw file](https://github.com/bblfsh/sdk/blob/481de6b05d4a247f5dafcb19a2905649452dbaa5/driver/manifest/discovery/discovery_test.go#L41)):*

Hmm, I still see this line in the PR, and I don't see the fix from the master here.
</blockquote></details>

I rebased now, sorry about that, I forgot to do it

ncordon:
___
*[uast/transformer/semantic.go, line 235 at r10](https://reviewable.io/reviews/bblfsh/sdk/434#-LrJPLLI-YakvQcAKL6J:-LrJi-fI705cumL1WqKu:b-e4fscb) ([raw file](https://github.com/bblfsh/sdk/blob/8fd1f1781b89cc078cb3af7b2f1d0257fb74d18a/uast/transformer/semantic.go#L235)):*

Please remove parentheses
</blockquote></details>

Done, I keep forgetting this because of C++ habits

ncordon:
___
*[uast/transformer/semantic.go, line 266 at r10](https://reviewable.io/reviews/bblfsh/sdk/434#-LrJPEYv03boBaSMsQma:-LrJi8ZmD-_-L7WIFW9h:bb6un1j) ([raw file](https://github.com/bblfsh/sdk/blob/8fd1f1781b89cc078cb3af7b2f1d0257fb74d18a/uast/transformer/semantic.go#L266)):*

Please remove parentheses
</blockquote></details>

Done

ncordon:
___
*[uast/transformer/semantic.go, line 289 at r10](https://reviewable.io/reviews/bblfsh/sdk/434#-LrJP6Lv6raqAQNDKiNY:-LrJiBV0Ce870BOdJfb2:b-rfolv7) ([raw file](https://github.com/bblfsh/sdk/blob/8fd1f1781b89cc078cb3af7b2f1d0257fb74d18a/uast/transformer/semantic.go#L289)):*

This is very slow and creates a lot of garbage! Please revert to `strings.Join`.
</blockquote></details>

I have used a `strings.Builder` instead

ncordon:
___
*[uast/transformer/semantic.go, line 198 at r8](https://reviewable.io/reviews/bblfsh/sdk/434#-LqkCngn2L1Mpw95Zuti:-LrKBeYaE3RS1DfTK2Yr:bq8v24k) ([raw file](https://github.com/bblfsh/sdk/blob/481de6b05d4a247f5dafcb19a2905649452dbaa5/uast/transformer/semantic.go#L198)):*

But you can still trim tabs with `DecodeRune`. Am I missing something?

This function in general takes a string and splits it into N parts (prefix, tab, lines, suffix). So it's definitely possible to use an original `string` value and slice it. This means that it's possible to implement it without any allocations by using `DecodeRune`. This won't require you to pass the `[]rune` slice around.

So do you want to try the `DecodeRune` approach, or should I review the current one? 
</blockquote></details>

Review the current one :pray: . This has been opened for a month and it is not such a big deal of fix IMHO

ncordon:
___
*[uast/transformer/semantic.go, line 277 at r11](https://reviewable.io/reviews/bblfsh/sdk/434#-LrOEY-p1J_K8guSw7O3:-LrOJPwKBEDlH1bNXB21:b-h6fomh) ([raw file](https://github.com/bblfsh/sdk/blob/11b28c9c59a036f421f53282dab206873df77e77/uast/transformer/semantic.go#L277)):*

The last bit that may affect performance: please move this line to the `if len(sub) == 0 {` below, so we won't create a large string that will become garbage immediately.
</blockquote></details>

I have changed it there

ncordon:
___
*[uast/transformer/semantic.go, line 278 at r12](https://reviewable.io/reviews/bblfsh/sdk/434#-LrOJTvgBPsGiKsOCv9n:-LrOJTvgBPsGiKsOCv9o:bci6236) ([raw file](https://github.com/bblfsh/sdk/blob/d4b06e63fb9883c3bcf5baa9a1b9dfbbb3135b19/uast/transformer/semantic.go#L278)):*
> ```Go
> 	var tab []rune
> 	// fast path, no tabs
> 	if len(sub) == 0 {
> ```

@dennwc I have also add it here since if we return here it means there is no common tab and we need to set the text and the tab from the comment

PR #443, opened 2019-10-29
Pool of native drivers (prototype)

### LINKED ISSUE #321: Prototype native parser scaling mechanism
Native parsers as processes should be also scalable (so far it's a single threaded, stdio process). This task is mainly about internal scalability mechanism.
The implementation should be preceded by some design document.
Also, the task will require changes in `sdk`, so it makes sense to attach a link to sdk's task, here.

This task is part of the epic: #313

kuba--:
Signed-off-by: kuba-- <kuba@sourced.tech>

This is the first draft/prototype which implements a pool of native drivers (parsers).
Instead of having just one native driver we handle a pool of concurrent drivers, so for each grpc.Parse request we can pull the driver, write request and put them back to the pool.
The pool of drivers was implemented as _buffered channel_.

This PR closes https://github.com/bblfsh/bblfshd/issues/321
and it's a part of epic: https://github.com/bblfsh/bblfshd/issues/313

Some design details you can find in doc: https://docs.google.com/document/d/1671ApmxFCOmerVKloVC6Q6j8BsJr8xZHbhuE40ZK-as/edit?pli=1#

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -32,6 +32,9 @@ var (
 
 	// ErrUnknownEncoding is returned for parse requests with a file content in a non-UTF8 encoding.
 	ErrUnknownEncoding = errors.NewKind("unknown source file encoding (expected UTF-8)")
+
+	// ErrDriverClosed is returned when Parse function was called after driver was closed.
+	ErrDriverClosed = errors.NewKind("driver has been closed")
 )
 
 // ErrMissingDriver indicates that a driver image for the given language
--- a/driver/impl.go
+++ b/driver/impl.go
@@ -10,14 +10,18 @@ import (
 	"github.com/bblfsh/sdk/v3/uast/nodes"
 )
 
-// NewDriver returns a new Driver instance based on the given ObjectToNode and list of transformers.
-func NewDriverFrom(d Native, m *manifest.Manifest, t Transforms) (DriverModule, error) {
-	if d == nil {
+// NewDriverFrom returns a new DriverModule instance based on
+// the given pool of native drivers, language manifest and list of transformers.
+func NewDriverFrom(ch chan Native, m *manifest.Manifest, t Transforms) (DriverModule, error) {
+	if ch == nil || len(ch) == 0 {
 		return nil, fmt.Errorf("no driver implementation")
-	} else if m == nil {
+	}
+
+	if m == nil {
 		return nil, fmt.Errorf("no manifest")
 	}
-	return &driverImpl{d: d, m: m, t: t}, nil
+
+	return &driverImpl{ch: ch, m: m, t: t, done: make(chan struct{})}, nil
 }
 
 // Driver implements a bblfsh driver, a driver is on charge of transforming a
@@ -25,21 +29,52 @@ func NewDriverFrom(d Native, m *manifest.Manifest, t Transforms) (DriverModule, 
 // `uast.ObjectToNode`` and a series of `tranformer.Transformer` are used.
 //
 // The `Parse` and `NativeParse` requests block the driver until the request is
-// done, since the communication with the native driver is a single-channel
-// synchronous communication over stdin/stdout.
+// done. The communication with the native driver is based on a buffer channel of drivers,
+// internally it allows number of concurrent requests equal to channel's size (by default - number of CPUs).
+// Communication with single driver is synchronous and goes over stdin/stdout.
 type driverImpl struct {
-	d Native
+	ch chan Native
 
-	m *manifest.Manifest
-	t Transforms
+	m    *manifest.Manifest
+	t    Transforms
+	done chan struct{}
 }
 
+// Start gets each driver from the channel, starts the process
+// and puts it back to the same buffer channel.
+// If any of processes fail on start,
+// the function tries to close all running drivers and return an error.
 func (d *driverImpl) Start() error {
-	return d.d.Start()
+	for i := 0; i < len(d.ch); i++ {
+		drv := <-d.ch
+		if err := drv.Start(); err != nil {
+			d.Close()
+			return err
+		}
+		d.ch <- drv
+	}
+	return nil
 }
 
+// Close tries to close all idle drivers.
+// If any of drivers fail on Close, the function returns the last received error.
 func (d *driverImpl) Close() error {
-	return d.d.Close()
+	select {
+	case <-d.done:
+		return nil
+
+	default:
+		close(d.done)
+
+		var last error
+		for i := 0; i < len(d.ch); i++ {
+			drv := <-d.ch
+			if err := drv.Close(); err != nil {
+				last = err
+			}
+		}
+		return last
+	}
 }
 
 // Parse process a protocol.ParseRequest, calling to the native driver. It a
@@ -52,25 +87,39 @@ func (d *driverImpl) Parse(rctx context.Context, src string, opts *ParseOptions)
 	if opts == nil {
 		opts = &ParseOptions{}
 	}
-	ast, err := d.d.Parse(ctx, src)
-	if err != nil {
-		if !ErrDriverFailure.Is(err) {
-			// all other errors are considered syntax errors
-			err = ErrSyntax.Wrap(err)
-		} else {
-			ast = nil
+
+	select {
+	case <-rctx.Done():
+		return nil, rctx.Err()
+
+	case <-d.done:
+		return nil, ErrDriverClosed.New()
+
+	case drv := <-d.ch: // get a native driver or wait on available one.
+		// put the driver back, when you're done.
+		defer func() { d.ch <- drv }()
+
+		ast, err := drv.Parse(ctx, src)
+		if err != nil {
+			if !ErrDriverFailure.Is(err) {
+				// all other errors are considered syntax errors
+				err = ErrSyntax.Wrap(err)
+			} else {
+				ast = nil
+			}
+			return ast, err
+		}
+		if opts.Language == "" {
+			opts.Language = d.m.Language
+		}
+
+		ast, err = d.t.Do(ctx, opts.Mode, src, ast)
+		if err != nil {
+			err = ErrTransformFailure.Wrap(err)
 		}
 		return ast, err
-	}
-	if opts.Language == "" {
-		opts.Language = d.m.Language
-	}
 
-	ast, err = d.t.Do(ctx, opts.Mode, src, ast)
-	if err != nil {
-		err = ErrTransformFailure.Wrap(err)
 	}
-	return ast, err
 }
 
 // Version returns driver version.
--- a/driver/server/common.go
+++ b/driver/server/common.go
@@ -1,32 +1,37 @@
 package server
 
 import (
+	"runtime"
+
 	"github.com/bblfsh/sdk/v3/driver"
 	"github.com/bblfsh/sdk/v3/driver/manifest"
 	"github.com/bblfsh/sdk/v3/driver/native"
 )
 
-var DefaultDriver driver.Native = native.NewDriver("")
-
-var (
-	// ManifestLocation location of the manifest file. Should not override
-	// this variable unless you know what are you doing.
-	ManifestLocation = driver.ManifestLocation
-)
+// ManifestLocation location of the manifest file. Should not override
+// this variable unless you know what are you doing.
+var ManifestLocation = driver.ManifestLocation
 
 // Run is a common main function used as an entry point for drivers.
-// It panics in case of an error.
+// It calls RunNative with buffered channel of default native drivers.
+// By default a size of the channel is equal number of CPUs.
 func Run(t driver.Transforms) {
-	RunNative(DefaultDriver, t)
+	n := runtime.NumCPU()
+	ch := make(chan driver.Native, n)
+	for i := 0; i < n; i++ {
+		ch <- native.NewDriver(native.UTF8)
+	}
+	RunNative(ch, t)
 }
 
-// RunNative is like Run but allows to provide a custom driver native driver implementation.
-func RunNative(d driver.Native, t driver.Transforms) {
+// RunNative creates and starts a new server for a given pool of native drivers.
+// It panics in case of an error.
+func RunNative(ch chan driver.Native, t driver.Transforms) {
 	m, err := manifest.Load(ManifestLocation)
 	if err != nil {
 		panic(err)
 	}
-	dr, err := driver.NewDriverFrom(d, m, t)
+	dr, err := driver.NewDriverFrom(ch, m, t)
 	if err != nil {
 		panic(err)
 	}
--- a/driver/server/server_test.go
+++ b/driver/server/server_test.go
@@ -23,7 +23,11 @@ func newDriver(path string) (*service, error) {
 	if err != nil {
 		return nil, err
 	}
-	d, err := driver.NewDriverFrom(native.NewDriverAt(path, native.UTF8), m, driver.Transforms{})
+
+	ch := make(chan driver.Native, 1)
+	ch <- native.NewDriverAt(path, native.UTF8)
+
+	d, err := driver.NewDriverFrom(ch, m, driver.Transforms{})
 	if err != nil {
 		return nil, err
 	}


bzz (commented):
> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)* status: 0 of 3 files reviewed, all discussions resolved

___
*[driver/server/common.go, line 19 at r1](https://reviewable.io/reviews/bblfsh/sdk/443#-LsQmuUh3VaJjfdJOPK3:-LsQmuUi2RjT2Uu-x0vm:b-wrgm75) ([raw file](https://github.com/bblfsh/sdk/blob/f3b6027d6df6a3d9ce5d22a954248c4e02ea7be9/driver/server/common.go#L19)):*
> ```Go
> func Run(t driver.Transforms) {
> 	n := runtime.NumCPU()
> 	ch := make(chan driver.Native, n)
> ```

Did you consider updating the go doc as well, to reflect this change?
This also seems like an opportunity to make sure all the exported functions are documented.

bzz (approved):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)* status: :shipit: complete! all files reviewed, all discussions resolved

kuba-- (commented):
> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[driver/server/common.go, line 19 at r1](https://reviewable.io/reviews/bblfsh/sdk/443#-LsQmuUh3VaJjfdJOPK3:-LsQv0qE5ELtj0dmx5PO:bshuhpu) ([raw file](https://github.com/bblfsh/sdk/blob/f3b6027d6df6a3d9ce5d22a954248c4e02ea7be9/driver/server/common.go#L19)):*
<details><summary><i>Previously, bzz (Alexander) wrote…</i></summary><blockquote>

Did you consider updating the go doc as well, to reflect this change?
This also seems like an opportunity to make sure all the exported functions are documented.
</blockquote></details>

Yeah, we'll have to  generally update API docs (it's on my todo). 
This PR is just to quickly show the workflow.

ncordon:
___
*[driver/impl.go, line 43 at r2](https://reviewable.io/reviews/bblfsh/sdk/443#-LsR_m5s8h4xJY8rAndt:-LsR_m5t8kUiKVBTe4Rp:b-25174u) ([raw file](https://github.com/bblfsh/sdk/blob/00043ffa19b27cc05fab3475bbcbf5a25042bb6f/driver/impl.go#L43)):*
> ```Go
> 
> // Start gets each driver from the channel, starts the process
> // and put it back to the same buffer channel.
> ```

Feel free to ignore this nitpick: *puts*

ncordon:
___
*[driver/impl.go, line 59 at r2](https://reviewable.io/reviews/bblfsh/sdk/443#-LsR_zyY5dK3pT6vfwVj:-LsR_zyY5dK3pT6vfwVk:b-4y0sib) ([raw file](https://github.com/bblfsh/sdk/blob/00043ffa19b27cc05fab3475bbcbf5a25042bb6f/driver/impl.go#L59)):*
> ```Go
> 
> // Close tries to close all idle drivers.
> // If any of drivers fail on Close, the function wraps an error (err) and move on.
> ```

*moves*

ncordon:
___
*[driver/impl.go, line 47 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRebKc9Mx9tMWUTNFA:-LsRebKd9L9RzvNVY5sR:b71xoht) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L47)):*
> ```Go
> // the function tries to close all running drivers and return an error.
> func (d *driverImpl) Start() error {
> 	for i := 0; i < len(d.ch); i++ {
> ```

Does this start all processes even if we are not going to use them? I mean if I have 8 processors, am I going to start 8 processes even if I am just going to parse a file?

ncordon (changes requested):
> Reviewed 1 of 3 files at r1, 1 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)* status: all files reviewed, 3 unresolved discussions (waiting on @kuba--)

kuba-- (commented):
> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)* status: all files reviewed, 5 unresolved discussions (waiting on @ncordon)

___
*[driver/impl.go, line 44 at r1](https://reviewable.io/reviews/bblfsh/sdk/443#-LsQvUTM4CmGNNxM4D9L:-LsQvUTM4CmGNNxM4D9M:b-zgkvib) ([raw file](https://github.com/bblfsh/sdk/blob/f3b6027d6df6a3d9ce5d22a954248c4e02ea7be9/driver/impl.go#L44)):*
> ```Go
> 		drv := <-d.ch
> 		if err := drv.Start(); err != nil {
> 			return err
> ```

Do you think it's OK to fail, if one of drivers failed on `Start`?
___
*[driver/impl.go, line 72 at r1](https://reviewable.io/reviews/bblfsh/sdk/443#-LsQvynx-hqKxVEvMh4Q:-LsQvynx-hqKxVEvMh4R:bfxgsak) ([raw file](https://github.com/bblfsh/sdk/blob/f3b6027d6df6a3d9ce5d22a954248c4e02ea7be9/driver/impl.go#L72)):*
> ```Go
> 	}
> 
> 	drv := <-d.ch
> ```

Is it ok, to block here? Maybe we may have some timeout, so if it'
___
*[driver/impl.go, line 43 at r2](https://reviewable.io/reviews/bblfsh/sdk/443#-LsR_m5s8h4xJY8rAndt:-LsReQ5wEPSGXsUX7q-S:b-896fix) ([raw file](https://github.com/bblfsh/sdk/blob/00043ffa19b27cc05fab3475bbcbf5a25042bb6f/driver/impl.go#L43)):*
<details><summary><i>Previously, ncordon (Nacho Cordón) wrote…</i></summary><blockquote>

Feel free to ignore this nitpick: *puts*
</blockquote></details>

Done.
___
*[driver/impl.go, line 59 at r2](https://reviewable.io/reviews/bblfsh/sdk/443#-LsR_zyY5dK3pT6vfwVj:-LsRePZoDp8KCAz57wob:b-896fix) ([raw file](https://github.com/bblfsh/sdk/blob/00043ffa19b27cc05fab3475bbcbf5a25042bb6f/driver/impl.go#L59)):*
<details><summary><i>Previously, ncordon (Nacho Cordón) wrote…</i></summary><blockquote>

*moves*
</blockquote></details>

Done.
___
*[driver/impl.go, line 47 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRebKc9Mx9tMWUTNFA:-LsRfVw697O_8cJfcmbL:b-j6zn54) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L47)):*
<details><summary><i>Previously, ncordon (Nacho Cordón) wrote…</i></summary><blockquote>

Does this start all processes even if we are not going to use them? I mean if I have 8 processors, am I going to start 8 processes even if I am just going to parse a file?
</blockquote></details>

Yes, it starts all parsers. Parsers will be _idle_ mode waiting for I/O.
It decreases latency. Otherwise, for instance for C++ driver (where parser is written in java), you will have to wait on JVM to start before parsing any file.

dennwc (changes requested):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)* status: all files reviewed, 8 unresolved discussions (waiting on @kuba-- and @ncordon)

___
*[driver/impl.go, line 44 at r1](https://reviewable.io/reviews/bblfsh/sdk/443#-LsQvUTM4CmGNNxM4D9L:-LsRlWpZ2K3QB59VrDm4:b5ksgx6) ([raw file](https://github.com/bblfsh/sdk/blob/f3b6027d6df6a3d9ce5d22a954248c4e02ea7be9/driver/impl.go#L44)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

Do you think it's OK to fail, if one of drivers failed on `Start`?
</blockquote></details>

Yeah, I think it's reasonable, at least for now. Later we can improve it if it becomes a problem.
___
*[driver/impl.go, line 72 at r1](https://reviewable.io/reviews/bblfsh/sdk/443#-LsQvynx-hqKxVEvMh4Q:-LsRli0F34FJMsm-kwtb:brzcmmg) ([raw file](https://github.com/bblfsh/sdk/blob/f3b6027d6df6a3d9ce5d22a954248c4e02ea7be9/driver/impl.go#L72)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

Is it ok, to block here? Maybe we may have some timeout, so if it'
</blockquote></details>

Yeah, you should `select` on 3 events: getting the driver, context cancellation, pool stop.
___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRksn-ErsR7TSkqAme:-LsRksn-ErsR7TSkqAmf:b2szd1o) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*
> ```Go
> // NewDriverFrom returns a new DriverModule instance based on
> // the given pool of native drivers, language manifest and list of transformers.
> func NewDriverFrom(ch chan Native, m *manifest.Manifest, t Transforms) (DriverModule, error) {
> ```

Accepting a channel is a very inconvenient API. Let's accept `func() Native` and a max number of drivers here instead.
___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRnDq7E3WXcwY6qDwI:-LsRnDq7E3WXcwY6qDwJ:bjvp19j) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*
> ```Go
> // NewDriverFrom returns a new DriverModule instance based on
> // the given pool of native drivers, language manifest and list of transformers.
> func NewDriverFrom(ch chan Native, m *manifest.Manifest, t Transforms) (DriverModule, error) {
> ```

Also, I think we need to implement scaling in the `native` package instead.

The implementation of `driver.Native` interface may accept multiple requests, and it does in case of Go driver. We don't need to scale that one, only ones created by `driver/native` package.
___
*[driver/impl.go, line 47 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRebKc9Mx9tMWUTNFA:-LsRl6pVB1l3-Lr_ko1r:bfaqxvc) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L47)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

Yes, it starts all parsers. Parsers will be _idle_ mode waiting for I/O.
It decreases latency. Otherwise, for instance for C++ driver (where parser is written in java), you will have to wait on JVM to start before parsing any file.
</blockquote></details>

@ncordon Also note that we do this to simplify things. In K8S we can disable the scaling here (set it to 1 driver) and let K8S scale whole drivers, not native ones. 
___
*[driver/impl.go, line 66 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRmCRBDNbRWFRPSWfY:-LsRmCRBDNbRWFRPSWfZ:bdjcq4n) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L66)):*
> ```Go
> 			// TODO(kuba--): replace following lines by errors.Wrap (after migrating to go 1.13)
> 			if err != nil {
> 				err = fmt.Errorf("%s, %s", err.Error(), e.Error())
> ```

No need to concatenate, just return the last error. Also, it may be better to make it explicit in the code (`var last error; for ... ; return last`).

kuba-- (commented):
> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)* status: all files reviewed, 7 unresolved discussions (waiting on @dennwc and @ncordon)

___
*[driver/impl.go, line 72 at r1](https://reviewable.io/reviews/bblfsh/sdk/443#-LsQvynx-hqKxVEvMh4Q:-LsSA_b_Bk-uyU5O4jOT:brk23p9) ([raw file](https://github.com/bblfsh/sdk/blob/f3b6027d6df6a3d9ce5d22a954248c4e02ea7be9/driver/impl.go#L72)):*

Yeah, you should `select` on 3 events: getting the driver, context cancellation, pool stop.
</blockquote></details>

do you mean to add `done` channel to the struct and do:
```
select {
case <- d.done:
    return nil, nil

case <- ctx.Done():
     return nil, ctx.Err()

case drv := <- d.ch:
//... al the code here
}
```

and in `Close` function do: `d.done <- struct{}{}`
This is what you mean?
I'm not sure if having and extra `done` channel makes sense. If the request already arrived on Close, I would let him go and return a response.
___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRksn-ErsR7TSkqAme:-LsS5yyT4FJjhQSxHKpL:b3wl3c1) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*

Accepting a channel is a very inconvenient API. Let's accept `func() Native` and a max number of drivers here instead.
</blockquote></details>

So how this function will work - like iterator?
How you gonna put them back, if you cannot deal with channel?

The option is to wrap this channel with a new structure `NativePool` with `Get`, `Put` and `New(size)` methods.
___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRnDq7E3WXcwY6qDwI:-LsS7petDyhEPx78W_WQ:b-64sfjg) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*

Also, I think we need to implement scaling in the `native` package instead.

The implementation of `driver.Native` interface may accept multiple requests, and it does in case of Go driver. We don't need to scale that one, only ones created by `driver/native` package.
</blockquote></details>

But the `native.go` represents one particular driver/parser - single process, so why scaling here?
And `impl.go` is just a implementation of particular grpc requests (Parse, Version, Languages, ...). If some day you want to scale `Parse2` or any other action then it's already there.
___
*[driver/impl.go, line 66 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRmCRBDNbRWFRPSWfY:-LsS8tv19UVA99gDIxnp:b-896fix) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L66)):*

No need to concatenate, just return the last error. Also, it may be better to make it explicit in the code (`var last error; for ... ; return last`).
</blockquote></details>

Done.

ncordon:
___
*[driver/impl.go, line 47 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRebKc9Mx9tMWUTNFA:-LsSEzGgCZiQR4-M5Dux:b-8y32dy) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L47)):*

@ncordon Also note that we do this to simplify things. In K8S we can disable the scaling here (set it to 1 driver) and let K8S scale whole drivers, not native ones. 
</blockquote></details>

:ok_hand: Thanks for the clarification

ncordon (approved):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)* status: all files reviewed, 5 unresolved discussions (waiting on @dennwc and @kuba--)

creachadair (commented):
Overall I think this is a good idea. I share Denys's concern about the package API, but the approach seems fine.

> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443#-:-LsSIqrX7F8YluqggiKq:b9ifr4k)* status: all files reviewed, 5 unresolved discussions (waiting on @dennwc and @ncordon)

___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRksn-ErsR7TSkqAme:-LsSHDsNEKW36bvxjiHG:b-gv9qxb) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

So how this function will work - like iterator?
How you gonna put them back, if you cannot deal with channel?

The option is to wrap this channel with a new structure `NativePool` with `Get`, `Put` and `New(size)` methods.
</blockquote></details>

It's fine to use a channel for the implementation, but I recommend passing in either a struct with callbacks or an interface. The problem with a channel in the API is that it's not clear how the caller is supposed to set up the channel, and what assumptions there are about the discipline for reading and writing it, except in documentation. With an interface the usage is more explicit.

Just a made-up example:

```go
type Pool interface {
    // Open blocks until a native driver is available, or until ctx terminates.
    Open(ctx context.Context) (Native, error)
    // Release adds or returns a driver to the pool.
    Release(Native)
}

// …

d, err := pool.Open(ctx)
if err != nil {
   doSomethingWith(err)
}
defer pool.Release(d)
doStuffWith(d)   
```

You could implement this with a channel, if you wanted.

```go
type pool chan Native

func (p pool) Open(ctx context.Context) (Native, error) {
    select {
    case v, ok := <-p:
        if ok {
            return p, nil
        }
        return nil, errors.New("pool closed")
    case <-ctx.Done():
        return nil, ctx.Err()
    }
}

func (p pool) Release(d Native) { p <- d }
```

But you would want to be careful probably to buffer the channel by at least 1 in either case, so that the pool does not deadlock on itself (this is true whether or not we wrap the type).
___
*[driver/impl.go, line 47 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRebKc9Mx9tMWUTNFA:-LsSIZng6ToQY8ct87Dm:b40p9og) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L47)):*
<details><summary><i>Previously, ncordon (Nacho Cordón) wrote…</i></summary><blockquote>

:ok_hand: Thanks for the clarification
</blockquote></details>

Also, idle processes are relatively cheap; as long as we're not giving them work, they will not chew up a lot of kernel resources and their resident sets can be paged out.

dennwc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)* status: all files reviewed, 5 unresolved discussions (waiting on @creachadair, @kuba--, and @ncordon)

___
*[driver/impl.go, line 72 at r1](https://reviewable.io/reviews/bblfsh/sdk/443#-LsQvynx-hqKxVEvMh4Q:-LsSkJMV4C4yt5r_XJ1T:b-vtax1p) ([raw file](https://github.com/bblfsh/sdk/blob/f3b6027d6df6a3d9ce5d22a954248c4e02ea7be9/driver/impl.go#L72)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

do you mean to add `done` channel to the struct and do:
```
select {
case <- d.done:
    return nil, nil

case <- ctx.Done():
     return nil, ctx.Err()

case drv := <- d.ch:
//... al the code here
}
```

and in `Close` function do: `d.done <- struct{}{}`
This is what you mean?
I'm not sure if having and extra `done` channel makes sense. If the request already arrived on Close, I would let him go and return a response.
</blockquote></details>

Yes, almost. Except that:

- Call `close(d.done)` in `Close` method. Also make sure it's safe to call it twice (`close` usually panics if you don't protect it).
- `return nil, ErrDriverClosed` when a signal from `d.done` is triggered.

To your point, you can't safely let the request go because `Close` will try to drain the channel and you may leak that driver that processes the request.
___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRnDq7E3WXcwY6qDwI:-LsSjNCV3HP8ZVNvALXQ:bxbp8ay) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

But the `native.go` represents one particular driver/parser - single process, so why scaling here?
And `impl.go` is just a implementation of particular grpc requests (Parse, Version, Languages, ...). If some day you want to scale `Parse2` or any other action then it's already there.
</blockquote></details>

You can make two separate struct types there - one will be with no scaling, and the second will scale the first one.

To your point, there is nothing in `Parse`/`Parse2` that says that it only allows one request at a time. So always scaling `driver.Native` doesn't sound right.

The `driver.Native` is not an interface for drivers implemented via stdin/stdout. It's an interface for any driver that returns a native AST. It may or may not be single threaded.
___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRksn-ErsR7TSkqAme:-LsSl6hRAfDX7OVrwnsD:bioa3nv) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*
<details><summary><i>Previously, creachadair (M. J. Fromberger) wrote…</i></summary><blockquote>

It's fine to use a channel for the implementation, but I recommend passing in either a struct with callbacks or an interface. The problem with a channel in the API is that it's not clear how the caller is supposed to set up the channel, and what assumptions there are about the discipline for reading and writing it, except in documentation. With an interface the usage is more explicit.

Just a made-up example:

```go
type Pool interface {
    // Open blocks until a native driver is available, or until ctx terminates.
    Open(ctx context.Context) (Native, error)
    // Release adds or returns a driver to the pool.
    Release(Native)
}

// …

d, err := pool.Open(ctx)
if err != nil {
   doSomethingWith(err)
}
defer pool.Release(d)
doStuffWith(d)   
```

You could implement this with a channel, if you wanted.

```go
type pool chan Native

func (p pool) Open(ctx context.Context) (Native, error) {
    select {
    case v, ok := <-p:
        if ok {
            return p, nil
        }
        return nil, errors.New("pool closed")
    case <-ctx.Done():
        return nil, ctx.Err()
    }
}

func (p pool) Release(d Native) { p <- d }
```

But you would want to be careful probably to buffer the channel by at least 1 in either case, so that the pool does not deadlock on itself (this is true whether or not we wrap the type).
</blockquote></details>

I think a single callback function should be enough here. `driver.Native` is already an interface and it has a `Start` and `Stop` methods that can get or put an actual implementation from the pool (if it has one).
___
*[driver/impl.go, line 66 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRmCRBDNbRWFRPSWfY:-LsSk8S3A3NJpVsrf-fT:b-g0pmk6) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L66)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

One more nit here: you can now call the error `err` instead of `e` (according to the convention).

creachadair (commented):
> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)* status: all files reviewed, 5 unresolved discussions (waiting on @dennwc, @kuba--, and @ncordon)

___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRksn-ErsR7TSkqAme:-LsT5g4OB5R8TSP1rUa9:b1bsk64) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*

I think a single callback function should be enough here. `driver.Native` is already an interface and it has a `Start` and `Stop` methods that can get or put an actual implementation from the pool (if it has one).
</blockquote></details>

That would also work, though then we would have to add another layer of wrapper for the existing implementations. I worry that might be a little too clever (and harder to debug).

I don't feel very strongly about which direction we go—my main point is to advise against using channels in the API definition, since it's very easy for the caller of the API to screw up the send/receive discipline and cause deadlocks or crashes. A pretty good rule of thumb for channels is they should always be internal details.

(I realize there are some counterexamples like `time.Ticker`, but even there you see some subtleties that people often get wrong around leaking goroutines and unfinished ticks)

kuba-- (commented):
> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)* status: 2 of 4 files reviewed, 5 unresolved discussions (waiting on @dennwc and @ncordon)

___
*[driver/impl.go, line 72 at r1](https://reviewable.io/reviews/bblfsh/sdk/443#-LsQvynx-hqKxVEvMh4Q:-LsTBfCd3TR5-5VZ_h8M:b-896fix) ([raw file](https://github.com/bblfsh/sdk/blob/f3b6027d6df6a3d9ce5d22a954248c4e02ea7be9/driver/impl.go#L72)):*

Yes, almost. Except that:

- Call `close(d.done)` in `Close` method. Also make sure it's safe to call it twice (`close` usually panics if you don't protect it).
- `return nil, ErrDriverClosed` when a signal from `d.done` is triggered.

To your point, you can't safely let the request go because `Close` will try to drain the channel and you may leak that driver that processes the request.
</blockquote></details>

Done.
___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRksn-ErsR7TSkqAme:-LsT1oQ87sbjw8qXN2Je:b55ay4q) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*

I think a single callback function should be enough here. `driver.Native` is already an interface and it has a `Start` and `Stop` methods that can get or put an actual implementation from the pool (if it has one).
</blockquote></details>

I would prefer to implement (on top of buffer channel)  sth. what Michael suggested.

Personally, I don't see the point of having just one `func` callback, and handle get and put by `Start` and `Stop`. `Start` and `Stop` for me are to start and stop the parser process. Both are called at the beginning and at the end.  In other words, having get/put implemented by start/stop means that we'll have to start and stop parser process for each Parse request.
___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRnDq7E3WXcwY6qDwI:-LsT7SQ912G5NTv0KluI:b-h814kd) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*

You can make two separate struct types there - one will be with no scaling, and the second will scale the first one.

To your point, there is nothing in `Parse`/`Parse2` that says that it only allows one request at a time. So always scaling `driver.Native` doesn't sound right.

The `driver.Native` is not an interface for drivers implemented via stdin/stdout. It's an interface for any driver that returns a native AST. It may or may not be single threaded.
</blockquote></details>

Honestly sounds a little bit overcomplicated. For me, this file is a pure implementation of grpc requests. So, we pull clients from the pool to talk to the backend (like usually happens with all DB stuff, but in this case the backend is running parser process and communication goes over stdin/out).

dennwc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @creachadair, @kuba--, and @ncordon)

___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRnDq7E3WXcwY6qDwI:-LsU-kqc-qJpzpAdBSBk:bqsj2cb) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

Honestly sounds a little bit overcomplicated. For me, this file is a pure implementation of grpc requests. So, we pull clients from the pool to talk to the backend (like usually happens with all DB stuff, but in this case the backend is running parser process and communication goes over stdin/out).
</blockquote></details>

I feel we are talking past each other.

Yes you are right, `driver.Native` is an analog of gRPC, but gRPC always assumes concurrency. The point I'm trying to make is that `driver.Native` (as an interface) is concurrent. And we have at least one implementation that is concurrent - Go driver.

So instead of stating that "every `driver.Native` needs a scaler to run concurrently", as your implementation implies, I suggested to hide this scaler behind the interface. And do this only for a single implementation that requires this kind of scaling - the one that runs over stdin/stdout (`driver/native.Driver`). It shouldn't be on a high level (`driver` package), since concurrency for other implementations can and will be implemented differently.

To illustrate the concept better, imagine Python driver that is implemented as a binding to C library of Python interpreter. Instead of starting N interpreters, and keeping them on hold you can just create/clone one on each `Parse` request, for example.

So the scaling you have implemented should be internal to a `driver/native` package, not external (in root `driver` package) as is in this PR.
___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRksn-ErsR7TSkqAme:-LsU18YB6QxQzhlMHMhm:b-vab37g) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

I would prefer to implement (on top of buffer channel)  sth. what Michael suggested.

Personally, I don't see the point of having just one `func` callback, and handle get and put by `Start` and `Stop`. `Start` and `Stop` for me are to start and stop the parser process. Both are called at the beginning and at the end.  In other words, having get/put implemented by start/stop means that we'll have to start and stop parser process for each Parse request.
</blockquote></details>

I think you heavily misunderstood me.

`Start` can wrap whatever is necessary for an implementation. In case of your implementation it will start N instances, and `Stop` will stop N instances. I don't understand why you say that `Start` should be called before each `Parse`.

I mentioned getting a driver from the pool because Michael mentioned that we may need an interface instead of a single callback. We don't, since `driver.Native` that is returned by the callback can use the pool internally, in it makes sense for an implementation. As an example I mentioned that `Start` may get N drivers from the pool, if we think that it's necessary.

So TL;DR is (IMO) that we already have an interface called `native.Driver`. No need for another wrapper on top of it.
___
*[driver/impl.go, line 62 at r5](https://reviewable.io/reviews/bblfsh/sdk/443#-LsU4pXJ-dijSgyVjJD4:-LsU4pXJ-dijSgyVjJD5:bnwvbox) ([raw file](https://github.com/bblfsh/sdk/blob/4d673c66ce186aedcf3004424636494d9b3b04a5/driver/impl.go#L62)):*
> ```Go
> // If any of drivers fail on Close, the function returns the last received error.
> func (d *driverImpl) Close() error {
> 	close(d.done)
> ```

```go
select {
case <-d.done:
default:
  close(d.done)
}
```
So it won't panic if called twice.

kuba-- (commented):
> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @dennwc and @ncordon)

___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRksn-ErsR7TSkqAme:-LsVwkVc-wcz3FfAmkQQ:b-ecmxa6) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*

I think you heavily misunderstood me.

`Start` can wrap whatever is necessary for an implementation. In case of your implementation it will start N instances, and `Stop` will stop N instances. I don't understand why you say that `Start` should be called before each `Parse`.

I mentioned getting a driver from the pool because Michael mentioned that we may need an interface instead of a single callback. We don't, since `driver.Native` that is returned by the callback can use the pool internally, in it makes sense for an implementation. As an example I mentioned that `Start` may get N drivers from the pool, if we think that it's necessary.

So TL;DR is (IMO) that we already have an interface called `native.Driver`. No need for another wrapper on top of it.
</blockquote></details>

Of course, if we implement it in such a way, it will work. But I wonder if the whole bootstrap won't be too heavy. Ok, right now  we hold all running processes, but thanks to this you get immediately the answer. If we create a process on demand we'll have to wait for JVM (C++ driver) or python interpreter (python driver) and so on..
Do you think, it's ok?
___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRnDq7E3WXcwY6qDwI:-LsW0iZUEhAmVe-Fe4jb:ba8zd72) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*

I feel we are talking past each other.

Yes you are right, `driver.Native` is an analog of gRPC, but gRPC always assumes concurrency. The point I'm trying to make is that `driver.Native` (as an interface) is concurrent. And we have at least one implementation that is concurrent - Go driver.

So instead of stating that "every `driver.Native` needs a scaler to run concurrently", as your implementation implies, I suggested to hide this scaler behind the interface. And do this only for a single implementation that requires this kind of scaling - the one that runs over stdin/stdout (`driver/native.Driver`). It shouldn't be on a high level (`driver` package), since concurrency for other implementations can and will be implemented differently.

To illustrate the concept better, imagine Python driver that is implemented as a binding to C library of Python interpreter. Instead of starting N interpreters, and keeping them on hold you can just create/clone one on each `Parse` request, for example.

So the scaling you have implemented should be internal to a `driver/native` package, not external (in root `driver` package) as is in this PR.
</blockquote></details>

OK, I got, your point. Of course we can simplify things here.
In this draft I had two goals:
- Don't change current API much, so it will be somehow transparent and  easy to read/understand (because of minimum number of changes).
- I  decided that if we have fixed number of concurrent processes  (instead of create them on demand), it's better to have them running and responsive (also look at my comment above).
___
*[driver/impl.go, line 62 at r5](https://reviewable.io/reviews/bblfsh/sdk/443#-LsU4pXJ-dijSgyVjJD4:-LsW59Mq5PVZVsmaFBTX:b-896fix) ([raw file](https://github.com/bblfsh/sdk/blob/4d673c66ce186aedcf3004424636494d9b3b04a5/driver/impl.go#L62)):*

```go
select {
case <-d.done:
default:
  close(d.done)
}
```
So it won't panic if called twice.
</blockquote></details>

Done.

kuba-- (commented):
> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @dennwc and @ncordon)

___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRnDq7E3WXcwY6qDwI:-LsWJX4o3Gw30FvGS9Eh:byfw4od) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

OK, I got, your point. Of course we can simplify things here.
In this draft I had two goals:
- Don't change current API much, so it will be somehow transparent and  easy to read/understand (because of minimum number of changes).
- I  decided that if we have fixed number of concurrent processes  (instead of create them on demand), it's better to have them running and responsive (also look at my comment above).
</blockquote></details>

One more thing. I think it's a drawback that we have sdk, but implement certain drivers differently (like go is concurrent, but others not). IMHO for distributed systems  consistent workflow is a key.
Anyway, current implementation does not force anyone to use any concurrency. You can still pass own pool (size 1) and do it by yourself.

Of course we can add another interface like:
```go
type Pool struct {
       Open() Native
       Close() error
}
```
make `DriverModule` as a composition of `Pool`, `Module`, `Driver`, 
and let drivers implement it, but it will require much more work across all drivers  and I wonder if we can really benefit from this flexibility.
Mainly because as we discussed this should be a tiny feature as a part of bigger refactoring, and any advanced scalability will be pushed to k8s.

dennwc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bblfsh/sdk/443)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @creachadair, @kuba--, and @ncordon)

___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRksn-ErsR7TSkqAme:-LsWpf-M9-oqyHxy-kgm:b-a11huc) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

Of course, if we implement it in such a way, it will work. But I wonder if the whole bootstrap won't be too heavy. Ok, right now  we hold all running processes, but thanks to this you get immediately the answer. If we create a process on demand we'll have to wait for JVM (C++ driver) or python interpreter (python driver) and so on..
Do you think, it's ok?

</blockquote></details>

Those are two different cases, and I specifically emphasized in both comments that the scaling is always implementation-dependent.

For Go driver you can just call the `Parse` function. `Start` and `Stop` do nothing.

For Python example, `Start` and `Stop` probably still do nothing and `Parse` may start  a new interpreter. Or maybe it gets those interpreters from the pool. It depends on the performance characteristics.

For JVM we can't embed it, so we use stdin/stdout, so we need additional  scaling. In this case `Start` creates N processes, `Parse` picks one of them.

For all 3 cases all the details are hidden behind `driver.Native` interface. That's the point I'm trying to make. For each implementation there may be a better way to achieve concurrency. For stdin/stdout we decided to start N processes. But it may not be the same for others.

So let's move the implementation where it belongs (`driver/native` package).

Also, to clarify, the example about Python is purely hypothetical. We won't implement it this way, it will still use stdin/stdout. But this is important to keep in mind for the future.
___
*[driver/impl.go, line 15 at r3](https://reviewable.io/reviews/bblfsh/sdk/443#-LsRnDq7E3WXcwY6qDwI:-LsWrLoM8WPUdUH4g3o5:b-t3qpmk) ([raw file](https://github.com/bblfsh/sdk/blob/195ef7746c3e97623f7b0fcf9c49f27c6f1ac4bd/driver/impl.go#L15)):*
> In this draft I had two goals:
> 
> Don't change current API much

But you are changing the API. This is exactly what I'm arguing about.

Go driver can run concurrently with no additional changes required. Your change, on the other hand, always adds scaling on top of it.

This is an API change because the expected behavior of the system changes. You are keeping `driver.Native` intact, but with a different set of assumptions: it's *must not* be concurrent and will be scaled *externally*. This restriction is not the case for a the current version of the `driver` package.

To clarify the last part, I'm speaking specifically about the `driver.Native` interface, e.g. if you use bblfshd as a library (which @kuba-- you were always an advocate for :)). bblfshd-the-app always scales drivers, but this is not because of the interface in the `driver` package, but rather because of how the app is designed.

So let's properly separate responsibilities for the implementations and not fall into the same trap as current bblfshd, which is to rely on a single implementation (libcontainer in case of bblfshd, and stdin/stdout as in case of this PR).

> I decided that if we have fixed number of concurrent processes

This is still a valid assumption and I was not arguing about this part here or in the other thread. Sorry if my comment to Michael confused you, when I mentioned the pool. It was solely to show that `driver.Native` is capable of hiding the process pool, scaling on demand or any other implementation detail.

So let's assume this topic is resolved. We are running exactly N processes, no scaling on demand for this PR, as we both agreed previously.

PR #445, opened 2019-11-12
add DriverFromClient constructor

lwsanty:
Signed-off-by: lwsanty <lwsanty@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/sdk/445)


protocol/driver.go
```diff
@@ -85,6 +85,13 @@ func AsDriver(cc *grpc.ClientConn) driver.Driver {
 	}
 }
 
+func AsDriverFromClient(c DriverClient, h DriverHostClient) driver.Driver {
```
kuba--: `DriverFromClient` would be good enough name IMHO

lwsanty: Done

